PR #13, opened 2023-03-06
refactor: use local files for genome resources 

deliaBlue:
fix: fix `test_cleanup.sh` to remove snakemake report  
refactor: change `prepare.smk`, `genome_process.sh` and `filter_anno_gtf.sh` to use local files for genome resources.


scripts/genome_process.sh
```diff
@@ -57,37 +51,22 @@ rm -fr "$logFile"; touch "$logFile"
...
-# Concatenate genome sequences fasta files
-echo "Concatenating genome sequences files..." >> "$logFile"
-genomeSeq="${resDir}/genome.fa.gz"
-for url in "${genomeSeqURLs[@]}"; do
-    cat "${rawDir}/$(basename "$url")" >> "$genomeSeq"
-done
-
-# Trim genome sequences IDs
+## TRIM GENOME SEQUENCES IDs
 echo "Triming genome sequenes IDs..." >> "$logFile"
```
uniqueg, line 55: I know you didn't do this and it isn't strictly a concern of _this_ issue, but there are two typos here, and given that you change a lot right before and after this line, you could also fix there here, I think:
- `Trimming` instead of `Triming`
- `sequence` instead of `sequenes`

scripts/genome_process.sh
```diff
@@ -57,37 +51,22 @@ rm -fr "$logFile"; touch "$logFile"
...
-
-# Concatenate genome sequences fasta files
-echo "Concatenating genome sequences files..." >> "$logFile"
-genomeSeq="${resDir}/genome.fa.gz"
-for url in "${genomeSeqURLs[@]}"; do
-    cat "${rawDir}/$(basename "$url")" >> "$genomeSeq"
-done
-
-# Trim genome sequences IDs
+## TRIM GENOME SEQUENCES IDs
```
uniqueg, line 54: Should be `## TRIM GENOME SEQUENCE IDs`, i.e., `SEQUENCE` in singular

scripts/genome_process.sh
```diff
@@ -57,37 +51,22 @@ rm -fr "$logFile"; touch "$logFile"
...
 echo "Triming genome sequenes IDs..." >> "$logFile"
+
 genomeSeqTrim="${resDir}/genome.processed.fa"
+cp "${genomeSeqFile}" "${output_dir}/genome.fa.gz"
+genomeSeq="${output_dir}/genome.fa.gz"
 gunzip "$genomeSeq"
-genomeSeq="${resDir}/genome.fa"
+genomeSeq="${output_dir}/genome.fa"
 awk '{if ($1 ~ /^>/) {print $1} else {print $0}}' "$genomeSeq" > "$genomeSeqTrim"
```
uniqueg, line 62: So all this script actually does after removing the download and concatenation is this line. I don't think we actually need a script for that and can just create a Snakemake rule `trim_genome_seq_ids` with that `awk` command in an Ubuntu container (instead of rule `genome_process`).

You'd only need to change the name of the rule, remove the `script` input, and replace:

```console
"(bash {input.script} {params.dir_out} {log} {input.genome})"
```

with something like:

```console
"(zcat {input.genome} | awk '{if ($1 ~ /^>/) {print $1} else {print $0}}' - > {output.genome) &> {log}"
```

(untested)

And of course you can remove the script `genome_process.sh` altogether.

scripts/genome_process.sh
uniqueg: Ignore this, see next comment.

scripts/genome_process.sh
uniqueg: Ignore this, see last comment in this file.

uniqueg (changes requested):
Apart from the in-line comments, the added test files are way too big. Please remove them.

Create small subsets of them, with no more than 10'000 lines each. We don't want to have files that are multiple megabytes in size under version control. If you have large test files (e.g., entire files for `chr19` or `chrY`), put them somewhere on our shared file system. For all test files, we should create test cases and run them automatically. For the small test files that we should include in the repo, we can do this with GitHub Actions. For larger tes cases, we should run them on our cluster.

For now, please just remove the test files you added here and please open three issues to deal with the tests and test files later:
1. Create a GitHub Actions continuous integration workflow
2. Prepare a workflow test case and include in GitHub Actions workflow (this should use _small_ files)
3. Prepare a full-size workflow test case and implement CI on cluster (this should use bigger, real-life test files like the ones you prepared for `chr19`)

scripts/filter_anno_gtf.sh
```diff
@@ -24,18 +24,16 @@ log_dir="$2"
 # Paths (DO NOT CHANGE!)  #Modified by Iborra P
 root="$PWD"
```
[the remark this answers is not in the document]
deliaBlue: If removing this step and pretend the user is providing an already filtered file, should we add a section in the `README.md` stating the assumptions we make for the input files? Maybe some recommendations or something like that?

deliaBlue:
> 
Then by now I have to provide a branch without any test files?

workflow/rules/prepare.smk
```diff
@@ -70,48 +63,28 @@ rule finish_prepare:
 
 
 ###############################################################################
-### Download and process genome IDs
+### Trimm genome IDs
```
uniqueg, line 66: "Trim", not "Trimm"

workflow/rules/prepare.smk
```diff
@@ -70,48 +63,28 @@ rule finish_prepare:
...
     input:
-        script=os.path.join(config["scripts_dir"], "genome_process.sh"),
+        genome=config["genome_file"],
     output:
         genome=os.path.join(
             config["output_dir"], "genome.processed.fa"
         ),
     params:
-        url=config["genome_url"],
         dir_out=config["output_dir"],
```
uniqueg, line 76: I think you're not using this one either anymore?

workflow/rules/prepare.smk
```diff
@@ -70,48 +63,28 @@ rule finish_prepare:
...
 ###############################################################################
-rule genome_process:
+rule trim_genome_seq_id:
     input:
-        script=os.path.join(config["scripts_dir"], "genome_process.sh"),
+        genome=config["genome_file"],
     output:
         genome=os.path.join(
             config["output_dir"], "genome.processed.fa"
         ),
```
uniqueg, line 74: While you are at it: this looks like it fits in one row with less than 80 chars.

uniqueg (changes requested):
Great job! Just a few very minor comments, shouldn't take more than 3 min :)

uniqueg (approved):

PR #14, opened 2023-03-06
fix(docs): Update dependencies

deliaBlue:
Upgrade `snakemake` to version `7.24.0` and `jinja2` to version `3.0`.  
Add `snakemake-minimal` version `7.24.0`.

### DIFF
--- a/environment.root.yml
+++ b/environment.root.yml
@@ -5,11 +5,11 @@ channels:
   - defaults
 dependencies:
   - graphviz=2.40.1
-  - jinja2=2.11.2
+  - jinja2=3.0
   - networkx=2.5
   - pygments=2.6.1
   - python=3.7.4
   - singularity=3.5.2
-  - snakemake=6.10.0
+  - snakemake=7.24.0
+  - snakemake-minimal=7.24.0
   - unzip=6.0
-  - wget==1.20.1
--- a/environment.yml
+++ b/environment.yml
@@ -5,10 +5,10 @@ channels:
   - defaults
 dependencies:
   - graphviz=2.40.1
-  - jinja2=2.11.2
+  - jinja2=3.0
   - networkx=2.5
   - pygments=2.6.1
   - python=3.7.4
-  - snakemake=6.10.0
+  - snakemake=7.24.0
+  - snakemake-minimal=7.24.0
   - unzip=6.0
-  - wget==1.20.1


environment.yml
```diff
@@ -5,10 +5,10 @@ channels:
...
   - graphviz=2.40.1
-  - jinja2=2.11.2
+  - jinja2=3.0
   - networkx=2.5
   - pygments=2.6.1
   - python=3.7.4
-  - snakemake=6.10.0
-  - unzip=6.0
-  - wget==1.20.1
+  - snakemake=7.24.0
```
uniqueg, line 12: Are _any_ of the dependencies, other than Snakemake and Python, actually _needed_? Did you try that out? I think that most of them would either be installed as part of Snakemake anyway, or they aren't actually needed (at least I wouldn't know where). Adding dependencies that are not needed (for either of these two reasons) just complicate our lives, because we will need to manage upgrading them much more often, and upgrades will also be more complicated.

Please create an environment with a recent version of Python, say `python=3.10`, from channel `conda-forge`. Then install Snakemake (from channel `bioconda`), without any version hint. Then see if your tests run. If yes, then add the Snakemake version that Conda actually installed to the environment file. If no, then try with a `snakemake=7.24.0` explicitly. If it still does not work, then you should hopefully get some error message that gives you a clue which additional dependency is really required. Add it, try again and so on. Only add what's needed to avoid errors.

Remove channel `defaults`.

environment.root.yml
```diff
@@ -5,11 +5,11 @@ channels:
   - defaults
 dependencies:
   - graphviz=2.40.1
-  - jinja2=2.11.2
+  - jinja2=3.0
   - networkx=2.5
   - pygments=2.6.1
   - python=3.7.4
   - singularity=3.5.2
```
uniqueg, line 12: Since you are working on the dependencies:

I don't like this pattern of repeating almost entire file contents. Sooner or later the dependencies in the two files will start to differ. Better to _only_ include Singularity in _this_ version of the file, then change the documentation (in the next PR that deals with the docs anyway) to tell people that if they want/need to install Singularity they should _update_ their environment with `environment.root.yml` rather than having users to choose one of the two files to _create_ the environment in the first place.

uniqueg: And note that since Singularity is taken from the `conda-forge` channel, you can remove `bioconda` and `defaults` channels. It will speed up the build and may also avoid compatibility issues (using different channels can lead to problems).

uniqueg (changes requested):
Apart from the in-line comments, the PR title has a number of issues:

- A `fix` refers to a _code change_ that changes the behavior of the software but does not change the API (in contrast to `feat` - which _does_ change the API). Using semantic versioning, it should lead to an increase in the patch version number, e.g., from `1.0.0` to `1.0.1`. In this PR, you are not changing the behavior of the software - in fact you are not even changing the code.
- As documentation is not code, pure doc changes can never be `fix` or `feat`, and also `fix(docs)` does not make sense (and anyway, this is not how the optional scope in Conventional Commits should be used). It makes even less sense here, because the dependencies are not part of the documentation, but part of the build system.
- Hence, you should use the `build` type (see [here](https://elixir-cloud-aai.github.io/guides/guide-contributor/workflow/#commit-messages) for the types we use)
- Unless using a proper noun, the description should start with a lower-case character

So please change to: `build: update dependencies`

environment.root.yml
deliaBlue: `bioconda` and `conda-forge` are both needed to install `snakemake`. Besides that, all the other dependencies can be removed.

deliaBlue: Also, in the `README.md` it is already stated to use `environment.root.yml` if `singularity` is not installed and the user has root permission.

uniqueg: Yes, but we want it to say that the users should install `environment.yml`, in any case, and then _update_ with `environment.root.yml` only if they want to install Singularity as well. In that case, `environment.root.yml` should _only_ contain Singularity, and nothing else.

uniqueg: And since `bioconda` is not needed to for Singularity (only `conda-forge`), you can remove `bioconda` from `environment.root.yml`.

environment.yml
```diff
@@ -1,14 +1,7 @@
 name: mirflowz
 channels:
-  - bioconda
   - conda-forge
-  - defaults
+  - bioconda
```
uniqueg, line 4: Put `bioconda` first - the order matters. And we want `Snakemake` from `bioconda`.

environment.root.yml
```diff
@@ -1,15 +1,8 @@
...
-  - jinja2=2.11.2
-  - networkx=2.5
-  - pygments=2.6.1
-  - python=3.7.4
+  - python=3.9.16
   - singularity=3.5.2
-  - snakemake=6.10.0
-  - unzip=6.0
-  - wget==1.20.1
+  - snakemake=7.24.0
```
uniqueg, line 8: Remove Snakemake and Python from this one, along with the `conda-forge` channel.

uniqueg (changes requested):

environment.root.yml
uniqueg: ...and of course you will need to update the `README.md` accordingly to reflect this new procedure (one-step if Singularity is already installed and two-step otherwise).

uniqueg (approved):

PR #15, opened 2023-03-07
docs: update documentation

deliaBlue:
Update the `README.md` to match the actual workflow.  
Remove images and add the one reflecting the actual workflow.  
Fix `test/test_rule_graph.sh` to provide the correct name to the rule graph.


uniqueg:
Please make sure you link all PRs to their corresponding issues. See [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) how to do that. In this case, you should add a line like this to the PR description:

```console
Resolves #11 
```

README.md
```diff
@@ -110,86 +110,122 @@ bash test/test_cleanup.sh
...
+gathering three different subworkflows,
+**_PREPARE_**, **_MAP_** and **_QUANTIFY_**. 
+Each one of them is implemented in an individual Snakemake file
+and fully containerized.    
+
 
 #### _PREPARE_
 
-The first workflow, **_PREPARE_** downloads and processes "genome resources"
+The first subworkflow, **_PREPARE_**, downloads and processes "genome resources"
```
uniqueg, line 129: Well, it does _not_ download anything after #13 is merged - and I thought the idea was to merge #13 and #14 first. We _can_ also merge this one first, and then you update the documentation again in #13 and #14. Your call. The important thing is that after merging _this_ PR, it has to be in sync with what is in the default channel, and everything coming in after that has to take care of its own changes to the doc, if necessary.

README.md
```diff
@@ -110,86 +110,122 @@ bash test/test_cleanup.sh
...
-But first, here is a brief description of what each of the three workflows
-does:
+But first, here is a brief description of what the workflow does:
 
 ### Workflow description
 
-The repository contains the following workflows, all implemented in Snakemake
-and fully containerized:
+The workflow consist on one main `Snakefile` 
+gathering three different subworkflows,
```
uniqueg, line 121: I actually don't think that users need to know that there are three subworkflows. It just makes things appear more complicated. Just merge the workflow description into one. And while you are at it, move it out of the "Usage" section - it doesn't make sense to be there. If I want to know how to use something, I don't want to be lectured on how it works 😆

README.md
```diff
@@ -110,86 +110,122 @@ bash test/test_cleanup.sh
...
+generated in the **_PREPARE_** subworkflow. Intersections are computed with
 [`bedtools`][bedtools] for one or multiple of mature, primary transcripts and
 isomiRs. Reads consistent with each miRNA are counted and tabulated.
 
-The scheme below is a visual representation of an example run of the
-**_QUANTIFY_** workflow:
 
-> ![rule-graph-quantify][rule-graph-quantify]
+The schema below is a visual representation of
+an example run of the whole workflow:
```
uniqueg, line 155: I know you didn't write this, but this is not correct. It _would_ be if you showed the workflow DAG instead of the rule graph. The rule graph is a representation of how the different rules are linked in a Snakemake workflow. It is a simplification of the DAG of an actual run (check the Snakemake docs to understand the difference).

README.md
```diff
@@ -110,86 +110,122 @@ bash test/test_cleanup.sh
...
-e.g., `MY_ANALYSIS`:
+```bash
+mkdir my_lib
+cd my_lib
+```
+Place your library file(s) here. 
+In addition, create a sample table. 
+Fill it with the correct entries. 
+You can look at the `test/test_files/sample_table.csv` 
+to get an idea of what this file must look like.
```
uniqueg, line 183: Users shouldn't only "get an idea" but be certain they understand the requirements of the input files. So apart from linking to the example, I think you should also describe the expected format (tab-separated, which columns are required...)

README.md
```diff
@@ -110,86 +110,122 @@ bash test/test_cleanup.sh
...
+or copy the script to run the workflow 
+on a **cluster via Slurm** 
+along with the `cluster.json` file with
+
+```bash
+cp ../test/test_workflow_slurm.sh ./run_workflow_slurm.sh
+cp ../test/cluster.json ./cluster.json
+```
+
+In both cases, the final 9 lines must be removed; this can be done with:
```
uniqueg, line 218: Remove the last 9 lines? 😆 

Instead of telling the users in detail that they should copy and then modify and trim the files by 9 lines (will you remember to update this part of the docs if you just add another two lines to the test script in 3 months?), just tell them what inputs are expected, where examples can be found, and how the Snakemake commands look like to:

- Start a dry run that creates a rule graph (can replace the section above)
- Start an actual run locally
- Start a run on an HPC with Slurm as example
- Create a report

And use different subsections for these.

uniqueg (changes requested):
Thanks. Much work to be done here. Apart from the in-line comments:

- I think you can remove `images/workflow_dag.svg`, I don't think it serves any useful purpose. It's also not referenced anywhere, as far as I'm aware
- You may need to modify the documentation describing how to create the workflow DAG, as well as the test script that creates them automatically
- To be sure it is not accidentally readded, you could also add it to `.gitignore`.

README.md
```diff
@@ -1,11 +1,11 @@
 # _MIRFLOWZ_
 
-Suite of [Snakemake][snakemake] workflows for the mapping and quantification
+_MIRFLOW_ is a [Snakemake][snakemake] workflow for the mapping and quantification
 of smallRNA-seq libraries, including miRNA and isomiR quantification.
```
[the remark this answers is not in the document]
deliaBlue: Should pre-miRNAs be included too?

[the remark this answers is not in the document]
uniqueg: Not in the summary, I think it's fine if you mention that in the longer description

README.md
deliaBlue: Should the workflow description be before the installation? bellow the summary?  I think it is better than have it in the middle or the bottom.

uniqueg: Both have advantages. Having it below the synopsis flows better, but people often want to get to the most important bits quickly, so synopsis > usage > description also makes sense. Up to you.

README.md
deliaBlue: Several things to finally get the sync documentation.

To start a local run, the user does not need to have a script, just to run the `snakemake` command, should I recommend to the command line on a script to be able to use it in different runs?  

Should the `test_cleanup.sh` should be recommended to be on the working directory in case some errors arise and must remove stuff?  

The creation of the report should be recommended to be added at the end of the run script so it is done altogether? This is, having a section on the `README.md` giving the command line to obtain a report but add the recommendation of having it altogether and provide an example of how the hole script should look like?

(still working on the cluster run, internet today is horribly slow)

uniqueg: You don't need to explain to people how Bash/shell works. If they want to put it in a script, they can do that (or find out elsewhere how, if they don't know already). Providing convenience scripts can be useful, but in this case, not all that much, because you restrict people's ability to change params easily. That's what a CLI is for. And we don't wanna start and build a CLI for _mirflowz_, it's not worth the trouble.

So just stick to the bare commands. The testing script is for running tests - that's more for _us_. You can include a reference on how to run that at the end of the installation section as an optional step to check whether all seems to have been installed/deployed fine. But I don't think it should be part of the _usage_ instructions.

README.md
```diff
@@ -48,20 +44,29 @@ conda env create -f environment.yml
 conda activate mirflowz
 ```
 
+If you have root permissions for your system and you do not already have
+`singularity` installed globally on your system, you must update the Conda
+environment using the `environment.root.yml` with command below. Mind that the
```
uniqueg, line 49: > "[...] with the command below."

_with_ "the" 

AND

> "Mind that you must have the [...]"

_without_ the first "the"

README.md
```diff
@@ -48,20 +44,29 @@ conda env create -f environment.yml
...
+you must have the environment activated to updated it.
```
uniqueg, line 50: "to update" (without the final `d`)

README.md
```diff
@@ -84,18 +89,12 @@ bash test/test_workflow_slurm.sh
...
 
-Execute the following commands to generate DAG and rule graph images for the
-workflow. The output will be found in the `images/` directory in the repository
-root.
-
-> **NOTE:** It is essential that you run the DAG and rule graph tests only
-> _after_ running the test workflow. This is because they require files to be
-> available that will only be created when running that workflows.
+Execute the following command to generate a rule graph image for the workflow.
+The output will be found in the `images/` directory in the repository root.
```
uniqueg, line 95: Good place to refer to the rule graph further below.

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
 
 ```bash
 mkdir MY_ANALYSIS
 cd MY_ANALYSIS
 ```
-Now within the `MY_ANALYSIS/` directory, 
-create a new directory to store you library; 
-name it as you like, e.g. `my_lib` and move there.    
+Place your library file(s) here. In addition, create the sample table 
+`sample_table.csv` and fill it up. The `sample_table.csv` is a tab-separated
```
uniqueg, line 124: "[...] and fill it in."

uniqueg, line 124: There's no reason to dictate a name for the sample table. They can call it `i_name_my_files_what_i_want.i_dont_care_about_file_extensions_either` and that's totally fine.

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
-> ![rule-graph][rule-graph]
-
-
-### Running the workflow
-
 Assuming that you are currently inside the repository's root directory, 
-create a directory from which you will run your workflow and
-name it whatever you want e.g., `MY_ANALYSIS` and 
-head to it.
+create a directory from which you will run your workflow and name it whatever
```
uniqueg, line 116: They don't need to do that. They can totally let their files stay where they are as long as Snakemake can access them.

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
 ```bash
 mkdir MY_ANALYSIS
 cd MY_ANALYSIS
 ```
-Now within the `MY_ANALYSIS/` directory, 
-create a new directory to store you library; 
-name it as you like, e.g. `my_lib` and move there.    
+Place your library file(s) here. In addition, create the sample table 
+`sample_table.csv` and fill it up. The `sample_table.csv` is a tab-separated  
+file, with one row per library, that must have the following columns:
```
uniqueg, line 125: For a better reading flow, you could put the high-level instruction first, then the code, and then some quote comment that adds details on what the sample table must look like. Something like this.

### Prepare inputs

1. Prepare sample table

```bash
touch path/to/you/sample/table.tab
```

> Fill in the sample table according to the following requirements:
> * foo
> * bar

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
+- `format`. In this field you mast state the library format. It can either be 
+`fa` if providing a FASTA file or `fastq` if the library is a FASTQ file.  
+
+You can look at the `test/test_files/sample_table.csv` to know what this file 
+must look like, or use it as a template.
 
 ```bash
 touch sample_table.csv
 ```
+In this same directory, add the following required input files:
```
uniqueg, line 140: They don't need to be in the same directory, they can be anywhere.

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
+must look like, or use it as a template.
 
 ```bash
 touch sample_table.csv
 ```
+In this same directory, add the following required input files:  
+
+- The gzipped reference genome in FASTA format.  
+- The gzipped gene annotation file in GTF format.  
+- The unzipped microRNA annotation file in GTF format.
```
uniqueg, line 144: Why "The". I don't think you mention them before. Also you could expand the description a bit. Add a section "Prepare genome resources" and then have a little paragraph for each one, including a link to Ensembl or miRBase.

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
 
-Now traverse back to the previous directory.
-Here, a couple of things must be done:  
+> **Note:** We expect the genome and gene annotations to be formatted according
+> the style used by Ensembl. Other formats are very likely to lead to problems.
+> The miRNA annotation file is expected to originate from miRBase, or follow 
+> their exact layout.
 
-First of all, copy the `config_template.yaml` to this directory.
+Finally, copy the `config_template.yaml` to this directory.
```
uniqueg, line 152: Again, put a separate subsection "Prepare configuration file". And again, it does not need to be in the same directory, it could be anywhere.

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
-```
 
-In both cases, the final 9 lines must be removed; this can be done with:
+With all the require files on the directory you can run the workflow with the
+following command:  
 
 ```bash
-head -n 9 run_workflow_*.sh
+snakemake \
+    --snakefile="../workflow/Snakefile" \
```
uniqueg, line 169: I would include the `--configfile` param to tell Snakemake where the config file is, and I would use generic paths, like `path/to/Snakefile` and `path/to/config.yaml`.

You could drop a hint that these paths don't need to be specifically provided if they live in certain places relative to the working directory (e.g., the docs explain where Snakemake looks for `--snakefile`: https://snakemake.readthedocs.io/en/stable/executing/cli.html)

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
+    --cores 4  \
+    --use-singularity \
+    --singularity-args "--bind ${PWD}/../" \
```
uniqueg, line 172: Why bind the directory upstream of your current working directory if you are in your repo root?
Anyway, this param probably needs an explanation, else people are bound to get into trouble.

README.md
```diff
@@ -48,20 +44,29 @@ conda env create -f environment.yml
...
 
+If you have root permissions for your system and you do not already have
+`singularity` installed globally on your system, you must update the Conda
+environment using the `environment.root.yml` with command below. Mind that the
+you must have the environment activated to updated it.
+
+```bash
+conda env update -f environment.root.yml
+```
+
```
uniqueg, line 55: I would put this command as a quote comment, like this:

> I'm a quote comment

README.md
```diff
@@ -1,11 +1,11 @@
 # _MIRFLOWZ_
 
-_MIRFLOW_ is a [Snakemake][snakemake] workflow for the mapping and quantification
-of smallRNA-seq libraries, including miRNA and isomiR quantification.
+_MIRFLOWZ_ is a [Snakemake][snakemake] workflow for mapping and 
+quantifying of miRNA and isomiR quantification.
```
uniqueg, line 5: Would be good to put a ToC here...

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
-```
 
-> **Note:** We expect the genome and gene annotations to be formatted according
-> the style used by Ensembl. Other formats are very likely to lead to problems.
-> The miRNA annotation file is expected to originate from
-> miRBase, or follow their exact layout.
+The schema below is a visual representation of the rules building up the 
+workflow and how they are related:
 
+> ![rule-graph][rule-graph]
```
uniqueg, line 231: Would be good to have sections on "Contributing", "License" and "Contact", such as here: https://github.com/elixir-cloud-aai/foca#contributing

Contact email is this: zavolab-biozentrum@unibas.ch

For contributing, just say something generic about welcoming everyone's contributions blabla. No code of conduct, not guidelines (unless you wanna deal with these, up to you).

uniqueg (changes requested):
Still plenty of things, but we're getting there :)

README.md
deliaBlue: Could it be more a suggestion than a mandatory step?

uniqueg: You could add a quote comment where you suggest that all inputs for a given run (or hard links pointing to them) are kept inside a dedicated directory, for example under the MIRFLOWZ root directory. And you can say that this makes it easer to keep the data together, reproduce an analysis and set up Singularity access to them. But do this only once at the beginning of the "Preparing inputs" section.

deliaBlue: I don't understand what do you mean by "set up Singularity access to them"...

README.md
deliaBlue: If we are specifying the path to the config file in the command line, should this same specification be commented/removed from the `snakemake` file?

README.md
uniqueg: Singularity can only access data from the file system if you explicitly mount a directory containing them via the `--bind my/directory` (relative or absolute). For example, if you mount the current working directory, Singularity can access all files in that directory, including in its subdirectories. Of course, in principle, you could just mount the file system's root directory with `/`, and Singularity would be able to access the entire file system. However, for security reasons it is strongly recommended to only mount the lowest level directory or directories necessary to access all input files.

Therefore, if your workflow requires multiple inputs and they are all spread across different paths of the file system, then you'd either need to mount multiple directories, or mount a single directory that is the parent of all the subdirectories containing the input files, but which may be on a higher level then necessary.

For example, assume your file system looks like this, you are currently in `/home/my_home/mirflowz` and the workflow requires access to files `input_?`:

```
/
    home/
        my_home/
            Documents/
                input_A
                Private/
                    # all kinds of private docs
            Downloads/
                input_B
            Secret/
                # all kinds of secret docs
            Work/
                mirflowz/
                    input_C
```

In this case you'd need to either mount `.` (may be mounted by default, not sure), `../../Documents` _and_ `../../Downloads`, or mount `../..` (or `/home/my_home` when using absolute paths).

Two issues with that:

1. In the first case your `--bind` command becomes more complicated, because you need to provide multiple arguments.
2. In both cases you'd give the container access to more data than necessary, e.g., the data under `/home/my_home/Documents/Private` (in both cases) and `/home/my_home/Secret` (in the second case).

If, however, all the inputs are under the current working directory, then you'd only need to mount `.` (or `/home/my_home/Work/mirflowz` when using absolute paths).

Does that make more sense now? Of course you don't need to describe this in detail. Just a small sentence, or half sentence, maybe with a link to the Singularity docs, will be enough)

P.S.: Check the test script to see how to pass the `--bind` parameter to Singularity via Snakemake.

README.md
deliaBlue: During the testing, the working directory is the `test/` therefore, the upstream directory is the repo root. I did not change it as I assumed that the user will create a directory to run the workflow thus, it will be the same as running it from the `test/` directory.

Should I change `${PWD}` for `src` and `../` for `dest` and give a couple of notes as well as referring the user to the documentation for more info?

README.md
```diff
@@ -113,244 +112,128 @@ bash test/test_cleanup.sh
...
-cp ../test/test_cleanup.sh ./run_cleanup.sh
-```
+### Running the workflow in an HPC with SLURM
 
-To start workflow execution, run:
 
-```bash
-./run_workflow_slurm.sh
-```
+TO BE FILLED
```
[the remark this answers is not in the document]
deliaBlue: Yesterday I found there are some problems with the workflow when running it from the cluster; should I create an issue and make the changes in another branch?

README.md
deliaBlue: Should I explain in this little paragraph how/for what will the file be used? I do not quite know what to write otherwise besides where to download the file. This is what I have rn: 

***

There are 4 files you must provide: 

- A gzipped reference genome in FASTA format. You can download it from [Ensembl][ensembl].  
- A gzipped gene annotation file in GTF format. You must provide the file with any filtering or modification desired already done. You can download it from the [Ensembl][ensembl] web page.
- An unzipped microRNA annotation file in GTF format. You can download this file from the [miRBase][mirbase] portal.
- A tab-separated mapping table between UCSC and Ensembl. You can find this tab-separated file for different organisms [here][chrMap].  

> **Note:** We expect the genome and gene annotations to be formatted according
> the style used by Ensembl. Other formats are very likely to lead to problems.
> The miRNA annotation file is expected to originate from miRBase, or follow 
> their exact layout.

README.md
uniqueg: Yes, good point. We should _not_ (never!) hardcode the path in the `Snakemake` file.

README.md
uniqueg: Ah yes, you are right. The best is to write scripts in a way that they can be executed from anywhere. But this is not a documentation issue, so you can open an issue for that and leave it as it is for this PR.

README.md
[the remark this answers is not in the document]
uniqueg: Yes. Leave the description of the cluster stuff out _completely_ for now (keep a copy somewhere of what you have already). Once you have the cluster test, then add the necessary documentation for that as well. This is anyway the proper way of doing things. You add a feature, you write/update the docs accordingly. We shouldn't have to write docs the way were are doing now (for all kinds of changes where we didn't adapt the docs), and we most certainly shouldn't write docs that will work some day in the future... maybe ;-)

README.md
deliaBlue: I quite don't get what the issue should look like ^^'

uniqueg: _Ensure that test scripts can be successfully run from any location_

uniqueg (approved):

PR #29, opened 2023-03-15
ci: fix broken test workflow

deliaBlue:


uniqueg (approved):
Difficult to comment on that. After all, the CI workflow passed on the PR branch before and only failed when merged. But will be happy to give it a shot.

deliaBlue:
> Difficult to comment on that. After all, the CI workflow passed on the PR branch before and only failed when merged. But will be happy to give it a shot once the tests pass.

It appears that when changing the file ucsc_to_ensembl.txt the results change. If only having the first entry of the file the counting all counts turn out to be 0; but if keeping all the list of chromosomes - even if they have no change -  the results are the ones expected. 
(By expected I mean that match the counting I made by hand).

I add all the entries again, change the expected output and make sure the library being used is the correct one.

uniqueg:
Okay, not quite sure I understand why this is the case, but I guess it's fine with me. Except that on the CI the tests still don't seem to work...

deliaBlue:
I made 5 local runs of the workflow and the results got me confused.
The first 2 and the last runs gave the expected output, but the runs 3 and 4 got 6 checksums - the same ones in the CI workflow - wrong. It seems like it is random weather these 6 files are generate in one way or another...  Maybe the rules that generate them need to be revisited and fixed

uniqueg:
So the inputs for that tool are the same in both cases, and then the outputs differ?

uniqueg:
Anyway, I think what we can do is create an issue for that and remove the problematic checksums for now. The results are the same, so it's not super urgent, and we shouldn't be blocked by this for too long.

uniqueg:
Presumably, multiple alignments for the same read are put in a hash / dictionary in that script. As these are unordered, when they are written out, the order can change between runs. So before writing out the remaining alignments for a record, a certain order should be enforced.

PR #32, opened 2023-03-17
feat: validate config file via JSON Schema

### LINKED ISSUE #9: Write a JSON schema for the config file
Write a JSON schema for the `config` file so that the optional parameters become truly optional and the `config` file is [validated by Snakemake](https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#validation) to be complete.

This would then be a valid MIRFLOWZ `config` file:

```yaml
samples: path/to/samples_table.csv
genome_url: path/to/gzipped/ensembl/genome.fa.gz
gtf_url: path/to/gzipped/ensembl/gene_annotations.gtf.gz
mirna_url: path/to/unzipped/mirbase/mirna_annotations.gff
map_chr_url: path/to/ucsc_ensembl_mappings.txt
```

deliaBlue:
Closes #9 .

### DIFF
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -28,7 +28,6 @@ from snakemake.utils import validate
 ### Configuration validation
 ###############################################################################
 
-configfile: "config.yaml"
 validate(config, os.path.join("..", "config", "config_schema.json"))
 
 
@@ -84,7 +83,7 @@ rule finish:
             ),
             sample=pd.unique(samples_table.index.values),
         ),
-        table1=expand(
+        table1=lambda wildcard:expand(
             os.path.join(
                 config["output_dir"], 
                 "TABLES", 
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -22,6 +22,14 @@
 
 import os
 import pandas as pd
+from snakemake.utils import validate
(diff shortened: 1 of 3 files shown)


workflow/Snakefile
```diff
@@ -22,6 +22,14 @@
 
 import os
 import pandas as pd
+from snakemake.utils import validate
+
+###############################################################################
+### Configuration validation
+###############################################################################
+
+configfile: "config.yaml"
```
uniqueg, line 31: But we don't want to hardcode the config file...

Btw: you can now remove all optional values from the config file (unless you wanna change the defaults) - so that should be part of this PR, too.

uniqueg (changes requested):
Did you test this with a valid and invalid config file? And please remove optional default params/values from the test `config.yaml`. This is also a good test.

deliaBlue:
> Did you test this with a valid and invalid config file? And please remove optional default params/values from the test `config.yaml`. This is also a good test.

I ran the test with the wrong  parameters two times and both of them raised en error. I left the `mir_list` option as by the moment, isomiRs are not being quantified.

uniqueg (approved):

PR #33, opened 2023-03-17
fix: stabilize alignment order

### LINKED ISSUE #30: State an order for multiple alignments for the same read
The script `sam_remove_duplicates_inferior_alignments_multimappers.1_5.pl` puts in a dictionary/hash  the different alignments for a single read . As these store data unordered the order may change between runs - which makes the test fail when comparing the output to the checksums. So before writing out the remaining alignments for a record, a certain order should be enforced.

deliaBlue:
Closes #30. 

Now the multiple alignments are first sort by the `POS` and in case of a tie, the `CIGAR` string is used.

Provided that the current `test_lib` does not have the read that was creating the conflict, 10 runs had been ran locally with a library containing that read. In all the runs, the same order has been observed.


uniqueg (approved):

uniqueg (changes requested):
Make sure you change the version of the script (to `1.6`), in the filename and in the usage subroutine (line 112), and in the Snakemake descriptor file.

uniqueg (approved):

PR #40, opened 2023-03-27
build: update tool versions

### LINKED ISSUE #37: Update all tools to recent versions.
Update all singularity tools' images to recent versions.

deliaBlue:
Closes #37


uniqueg:
Can you please post the `segemehl` error and some lines of the original output of `bedops` and that you get with a recent version?

Did you try to look at the changelogs of the tools to try to understand why you get these differences? Are there perhaps any changes in the CLI params (new/changed) that could explain these differences? Perhaps small changes in the commands would avoid the problems.

But if you don't find anything, it's okay to keep `segemehl` and `bedops` at their current versions. Or better would be to identify an acceptable range of versions that give us the output we expect. For example `segemehl` might need to be at least `0.2` but smaller than `0.3`.

deliaBlue:
I looked at the different manifest layers to try spot any change but it seems I did not have to be looking there. I search how to see the changelogs and it seems it is done with the `docker history` command but again, could not find anything (at least understandable for me). 

## Bedops

The `bedops` actual version output (`mirna_filtered.bed`)  is:

```
19	38	122	MI0003140	.	+	.	miRNA_primary_transcript	.	ID=MI0003140;Alias=MI0003140;Name=hsa-mir-512-1
19	51	74	MIMAT0002822	.	+	.	miRNA	.	ID=MIMAT0002822;Alias=MIMAT0002822;Name=hsa-miR-512-5p;Derives_from=MI0003140
19	88	110	MIMAT0002823	.	+	.	miRNA	.	ID=MIMAT0002823;Alias=MIMAT0002823;Name=hsa-miR-512-3p;Derives_from=MI0003140
```

The `bedops` updated version is:

```
19	38	122	.	.	+	.	miRNA_primary_transcript	.	ID=MI0003140;Alias=MI0003140;Name=hsa-mir-512-1
19	51	74	.	.	+	.	miRNA	.	ID=MIMAT0002822;Alias=MIMAT0002822;Name=hsa-miR-512-5p;Derives_from=MI0003140
19	88	110	.	.	+	.	miRNA	.	ID=MIMAT0002823;Alias=MIMAT0002823;Name=hsa-miR-512-3p;Derives_from=MI0003140
```

## Segemehl

The error that raises is found in the rule `remove_inferiors`:

```
[ERROR] Edit distance ('NM tag') missing! at ../scripts/sam_remove_duplicates_inferior_alignments_multimappers.1_6.pl line 185, <IN> line 5.
```

And in fact, some others besides the one in line 5 are missing the "NM tag" when using the newest version of `segemehl` on the `map` rule (there's no problem of updating `segemehl` in the `prepare` rule but for consistency if one is not updated the other shouldn't either)

uniqueg:
The point for checking the changelogs was to identify what changed between versions and how we could update the actual commands to give us the desired outcomes. For example, there may, in never versions, be command line options that give us the NM tag, but that now need to be specifically provided. Similarly, there may be an option to modify the `bedops` output such that the result is consistent with what we got with previous options.

deliaBlue:
I've check the `bedops` changelog; apparently the newest version puts by default the gene ID on column 4. With the argument `--attribute-key` you can select what you want in the fourth column. I tried to explicitly request the gene ID, I tried with the `transcript_id` too and both times I got the fourth column with nothing but `.`.

Regarding `segemehl`, the changes made for the new version tell nothing about the `NM` tag. Not even in the docu they give any command to modify the tags nor anything similar.

uniqueg:
Maybe you should file a bug report against `bedops` then, as this sounds very much like a bug (unless you did something wrong, in which case they may want to improve their docs). In both cases this is on their end, and until its resolved or clarified, please keep the old version.

As for `segemehl`: Did you try to call it with the `--help` option or read its documentation to see if you find any evidence of a CLI param that may include the `NM` tag in the output?

uniqueg:
It seems that `segemehl` should still output the `NM` tag: https://www.bioinf.uni-leipzig.de/Software/segemehl/

If it doesn't, it is, again, either a bug, or the `segemehl` command or the follow up command running into a problem may need to be modified.

deliaBlue:
> Maybe you should file a bug report against `bedops` then, as this sounds very much like a bug (unless you did something wrong, in which case they may want to improve their docs). In both cases this is on their end, and until its resolved or clarified, please keep the old version.
> 
> As for `segemehl`: Did you try to call it with the `--help` option or read its documentation to see if you find any evidence of a CLI param that may include the `NM` tag in the output?

On their changelog they state to have fixed this exact bug (at least for gtf2bed). Anyways, I'll keep the actual version.

deliaBlue:
> It seems that `segemehl` should still output the `NM` tag: https://www.bioinf.uni-leipzig.de/Software/segemehl/
> 
> If it doesn't, it is, again, either a bug, or the `segemehl` command or the follow up command running into a problem may need to be modified.

I found the problem with the output of the `segemehl` file. It turns out that now `segemehl` also returns the unmapped reads. They provide the parameter `-u` to drop the unmapped sequences there but the final file still contain these "empty alignments". The solution, I believe, is to create a new rule that processes the output file and removes the unaligned sequences. Should it be done in this same PR?

uniqueg:
Yes, I agree with the extended CIGAR being potentially useful. And I also think it should go in another PR, keeping `v0.2` in this one.

As for `bedops`: even more reason to open an issue or add on to the existing one. But yes, keep the version for now.

uniqueg (approved):

uniqueg:
Yes, I agree with the extended CIGAR being potentially useful. And I also think it should go in another PR, keeping `v0.2` in this one.

As for `bedops`: even more reason to open an issue or add on to the existing one. But yes, keep the version for now.

uniqueg:
Please resolve the conflicts so that I can merge this PR.

uniqueg (dismissed):

deliaBlue:
With the merge with the `dev` branch the expected output must be changed as now miRNA annotations are not filtered (I resolved the conflict the other way around).

uniqueg (approved):

deliaBlue:
> As for `bedops`: even more reason to open an issue or add on to an existing one. But yes, keep the version for now. Please link to the issue discussing the bug here (if there is one), so that we can find it back later on.

Here's the link to the [issue](https://github.com/bedops/bedops/issues/244#issuecomment-1489764549). I already comment on it

PR #41, opened 2023-03-29
refactor:  extending pre-miR overhangs

### LINKED ISSUE #24: Extending pre-miRs overhang
The 5p and 3p pre-miR overhang relative to the canonical miRNA has to be extended up to $n$ if it is smaller than $n$.  
$n$ will be left for the user to choose and will have a default value of 10 which - by the moment will also be the maximum allowed value.  

> MIND that if any pre-miR starts at a position $< n$ or ends at a position  $> lenSeq - n$ the extension will be either as big as the difference to the  chromosome ends or there will be no extension.

deliaBlue:
Closes #24 .


scripts/extend_overhang.py
```diff
@@ -0,0 +1,155 @@
...
+
+with open(args.chr, "r") as f:
+    chr_size = int(f.readline().split()[1])
+
+extension =  args.extension
+
+
+def extend_overhang(pre, mirna_start, mirna_end, extension, chr_size):
+    '''
+    EXTEND OVERHANG FUNCTION
```
uniqueg, line 70: Please use Google-style docstrings, and make sure to provide them for every module, class, method and function.

Also, please provide type hints at least for all function and method signatures, as well as for global variables (which you should anyway avoid, if at all possible). Also consider adding them for local variables, because, why not. Better to be safe.

scripts/extend_overhang.py
```diff
@@ -0,0 +1,155 @@
...
+    mirna_end: highest position of the mature form(s)
+    extension: Minimum number of basepairs to have as overhang 
+    chr_size: Size of the sequences chromosome
+
+    OUTPUT
+    ---------
+    pre: pre-miRNA modified entry 
+    '''
+
+    over_1 = int(mirna_start) - int(pre[3])
```
uniqueg, line 89: You're making use of indices too much. It's dangerous and hard to read/maintain. Why not use a GFF parser. Maybe there's one in Biopython?

uniqueg (changes requested):
Plenty of work. I didn't even review the logic. Will have a look again when my comments are addressed and the code is nice and clean :)

config/config_schema.json
```diff
@@ -45,15 +45,10 @@
...
-            "description": "Number of shifts relative to the start of each annotated mature miRNA."
-        },
-        "bp_3p":{
-            "type": "array",
-            "default": [-2, -1, 0, 1, 2],
-            "description": "Number of shifts relative to the end of each annotated mature miRNA."
+        "overhang":{
+            "type": "integer",
+            "default": 10,
+            "description": "Minimum length (in bp) of the pre-miRs overhang. Its value cannot exceed 10 bp."
```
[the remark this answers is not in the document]
deliaBlue: In VS code the indentation is the same as the other params.

deliaBlue:
The output was update due to two reasons:
- The miRNA annotations went back to `Derives_from` therefore, the intersection file contains `Derives_from` instead of `Parent`.
- While testing, I found an annotation out of the genome range, it was removed. This makes the tables to have one less entry but the same counting thus, they were updated.

uniqueg (commented):
I guess we leave this until after you have made that plan and confirmed that our strategy, as discussed yesterday, is good

PR #43, opened 2023-04-02
refactor: filter multimappers

deliaBlue:


scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+def filter_multimappers(in_sam_file, out_sam_file):
+    '''
+    Filters multimappers from a SAM file and writes the filtered reads to a new
+    SAM file.
+
+    This function scans a SAM file recording how many alignments per read there
+    are. Then, for the reads with multiple alignments (multimappers), it counts
+    the number of matches and indels it contain. Finally, it writes to a new SAM
+    file the alignments with the more matches and the less indels. If different 
+    alignments obtain the same score, all of them will be kept.
```
uniqueg, line 74: I think you can move this to the top as a module level docstring (see comments above and below). However, I don't think it's very well written. Perhaps you can ask ChatGPT to explain your script to you and then use that (following the suggested summary line from above) :)

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Args:
+        in_sam_file (str): 
+            Path to the input SAM file.
+        out_sam_file (str): 
+            Path to the output SAM file.
+
+    Returns:
+        None
```
uniqueg, line 83: You can skip `Returns` if you return `None`.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+def count_indels(read):
+    '''
+    Counts the number of insertions and/or deletions in a read based on its 
+    CIGAR string.
+
+    Args:
+        read (pysam.AlignedSegment): The read to count insertions for.
+
+    Returns:
+        int: The number of insertions in the read.
```
uniqueg, line 61: Rather: The _sum_ of insertions _and deletions_ in the _alignment_

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+def count_matches(read):
+    '''
+    Counts the number of matches in a read based on its CIGAR string.
+
+    Args:
+        read (pysam.AlignedSegment): 
+            The read to count matches for.
+
+    Returns:
+        int: The number of matches in the read.
```
uniqueg, line 47: In the _alignment_

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+    return parser
+
+
+def count_matches(read):
+    '''
+    Counts the number of matches in a read based on its CIGAR string.
+
+    Args:
+        read (pysam.AlignedSegment):
```
uniqueg, line 43: It's not really a read, it's an alignment (of a read against a reference)

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+        type=str,
+        default=None,
+        )
+
+    parser.add_argument(
+        '-o', '--output', 
+        help="Path for the gff3 final file.",
+        type=str,
+        default=None,
+        )
```
uniqueg, line 34: Again, as in the other script: Better to write to `STDOUT` (`pysam` supports that).

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+    parser.add_argument(
+        '-v','--version', 
+        action='version', 
+        version='%(prog)s 1.0',
+        help="Show program's version number and exit"
+        )
+
+    parser.add_argument(
+        '-i', '--input', 
+        help="Path to the gff3 input file",
```
uniqueg, line 24: Sure it's gff3? I think it's SAM, no?

"Path to input file in SAM format, sorted by query name"

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+    read_dict = {}
+    for read in samfile:
+        # Skip secondary and supplementary alignments
+        if read.is_secondary or read.is_supplementary:
+            continue
+        
+        if read.query_name in read_dict:
+            read_dict[read.query_name].append(read)
+        else:
+            read_dict[read.query_name] = [read]
```
uniqueg, line 100: Not the best idea to build a dictionary from an entire SAM file, which may be gigabytes in size. All of that stuff needs to stay in memory then. Better to go read by read.

You just need to add the assumption that reads are sorted by query name (which is fine in our case anyway, as the reads are just coming out of the previous filtering step, which surely uses the same approach).

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # Iterate over reads with the same query name
+    for read_name in read_dict:
+        reads = read_dict[read_name]
```
uniqueg, line 104: `for alignments in read_dict.values()` would be sufficient here. Or `for read_name, alignments in read_dict.iterate()` to have both keys (which you don't appear to need) and values.

But this will change anyway when you don't put all of your alignments in a dict.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+        out_sam_file (str): 
+            Path to the output SAM file.
+
+    Returns:
+        None
+    '''
+
+    # Open input SAM file and create output SAM file
+    samfile = pysam.AlignmentFile(in_sam_file, "r")
+    out_sam = pysam.AlignmentFile(out_sam_file, "w", template=samfile)
```
uniqueg, line 88: Consider using context managers.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+        reads = read_dict[read_name]
+        if len(reads) == 1:  
+            out_sam.write(reads[0])
+        else:  
+            # Create list of tuples with read object, number of matches and indels
+            read_counts = [(r, count_matches(r), count_indels(r)) for r in reads]
+            # Find the maximum counts for matches and indels
+            max_counts = max(read_counts, key=lambda x: (x[1], -x[2]))[1:]
+            # Create list of all reads with the same maximum counts
+            best_reads = [mr[0] for mr in read_counts if mr[1:] == max_counts]
```
uniqueg, line 113: I would probably put this stuff in a separate function.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,133 @@
...
+            read_dict[read.query_name] = [read]
+
+    # Iterate over reads with the same query name
+    for read_name in read_dict:
+        reads = read_dict[read_name]
+        if len(reads) == 1:  
+            out_sam.write(reads[0])
+        else:  
+            # Create list of tuples with read object, number of matches and indels
+            read_counts = [(r, count_matches(r), count_indels(r)) for r in reads]
```
uniqueg, line 109: I don't quite understand why you need both indels and matches. Given that the assumption is that multimappers all have the same edit distance (you should mention that assumption in the module/script docstring, along with the requirement for the SAM file to be sorted by query name), wouldn't any _one_ of the two be sufficient?

This would very much simplify this code and have better performance.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+
+Usage:
+    python filter_multimappers.py -i input_sam_file
+    python filter_multimappers.py -i input_sam_file > output_sam_file
+"""
+from pathlib import Path
+from typing import List
+import argparse
+import pysam
+import sys
```
uniqueg, line 44: Have a look at [PEP 8](https://peps.python.org/pep-0008/) to see how to order imports into different sections. We also like to sort each section alphabetically by the name of the package, so `from c import d` would come after `import b`, even though `import` comes before `from`, alphabetically.

To help with the sections and sorting, you can use `isort` (https://pycqa.github.io/isort/index.html), with setting `--force-sort-within-sections` to `True`: https://pycqa.github.io/isort/docs/configuration/options.html#force-sort-within-sections

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+If the current query changes, it selects the best alignment(s) from the list
+of current alingments and writes them to the standard output. Current query
+and current alignments list are reset to the new query and alignment
+respectively.
+Finally, if there are any remaining alignments in the current alignments
+list and selects the best alignment(s) for those queries, writing them to
+standard output.
+
+Input:
+    input_sam_file (str): Path to the input SAM file, ordered by query name.
```
uniqueg, line 19: Every CLI argument is a string. In fact, `str` doesn't even make sense for the shell, so you can remove that here. Rather, usage strings typically use metavars like `FILE` or `PATH`.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+    find_best_alignments(list[pysam.AlignedSegment]):
+        Find alignments with less indels
+    write_output(list[pysam.AlignedSegment]):
+        Write the output to the standard output (stdout).
+    main(Path):
+        Filter multimappers by indels count.
+
+Usage:
+    python filter_multimappers.py -i input_sam_file
+    python filter_multimappers.py -i input_sam_file > output_sam_file
```
uniqueg, line 38: Given that you put a shebang in the first line, if you make the script executable, you don't need to include `python` here. Also, use the metavar like `FILE` or `PATH` or `SAM` instead of `input_sam_file`:

```bash
filter_multimappers.py SAM
```

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+    """Count the number of indels in an alignment based on its CIGAR string.
+
+    This function counts the number of indels in the alignment based on the
+    alignment CIGAR string returned by .cigartuples. Insertions are encoded as
+    1s whereas the deletions are encoded as 2.
+    Refer to the .cigartuples documentation for more information at:
+    https://pysam.readthedocs.io/en/v0.15.0/api.html#pysam.AlignedSegment.cigartuples
+
+    Args:
+        alignments(pysam.AlignedSegment): The read to count insertions for.
```
uniqueg, line 79: No types in the docstrings please. We always want to avoid situations where we have multiple sources of truth. Otherwise, the risk is too high that one, but not the other part gets updated. If possible, it's always better to have the code be the single source of truth. For the same reason, we want to have as few inline comments as possible. Use them only in really hard to understand parts of your code. And if you _have_ hard to understand parts in your code, _first_ think whether you can make them easier to understand. And _only_ if that is not possible (or would be much more costly), then add an inline comment.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+    """Filter multimappers by indels count.
+
+    Args:
+        sam_file (str): Path to the input SAM file.
+    """
+    with pysam.AlignmentFile(sam_file, "r") as samfile:
+
+        # Extract header from input SAM file and write it to stdout
+        header = samfile.header
+        sys.stdout.write(str(header))
```
uniqueg, line 140: This is not a hard to understand code area, so no comment required (see my comment above). In fact, the code is just as easy to read as your comment. It would still be as easy to read (or even easier) and actually _shorter_ than your prose comment, if you put it all in one line:

```python
sys.stdout.write(str(samfile.header))
```

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+                current_alignments = [alignment]
+            else:
+                current_alignments.append(alignment)
+
+        # Last set of alignments
+        if len(current_alignments) == 1:
+            write_output(alignments=current_alignments)
+        else:
+            best_alignments = find_best_alignments(alignments=current_alignments)
+            write_output(alignments=best_alignments)
```
uniqueg, line 172: This part occurs twice, so best to put it in a function. Makes the `main()` function cleaner as well. As a general rule, you should try to keep a function below a dozen statements or so, if possible (of course there are many good reasons not to, so it's certainly not a strict rule).

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+                    best_alignments = find_best_alignments(alignments=current_alignments)
+                    write_output(alignments=best_alignments)
+
+                # Reset variables
+                current_query = alignment.query_name
+                current_alignments = [alignment]
+            else:
+                current_alignments.append(alignment)
+
+        # Last set of alignments
```
uniqueg, line 167: Remove all of these comments, none of it is hard to read/understand.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+                    write_output(alignments=current_alignments)
+                else:
+                    best_alignments = find_best_alignments(alignments=current_alignments)
+                    write_output(alignments=best_alignments)
+
+                # Reset variables
+                current_query = alignment.query_name
+                current_alignments = [alignment]
+            else:
+                current_alignments.append(alignment)
```
uniqueg, line 165: I would switch the `if` and `else`, for two reasons:
1. The appending of alignmets of the same read/query happens _before_ their processing/filtering, so it makes sense for the code where you append those alignments to come first.
2. It is easier to parse `==` compared to `!=`, so if there is no good reason to use negatives (and in fact, there is rather another reason _not_ to use the negative, see reason above), then don't use them.

uniqueg (changes requested):
I think this will be last round of comments. Sorry for being anal here, but you're here to learn better coding (among other things). Also, with that extra workin in the beginning, reviewing should become easier with time (both for me and you) while still having high quality code.

scripts/filter_multimappers.py
```diff
@@ -0,0 +1,177 @@
...
+    Retrns:
+        best_alignments (list[pysam.AlignedSegment]): alignments with the less indels
+    """
+    alignment_indels = [(aln, count_indels(alignment=aln)) for aln in alignments]
+    min_indels = min(alignment_indels, key=lambda x: x[1])[1]
+    best_alignments = [indelCount[0] for indelCount in alignment_indels if indelCount[1] == min_indels]
+    
+    for i in range(len(best_alignments)):
+        best_alignments[i].set_tag('NH', len(best_alignments))
+        best_alignments[i].set_tag('XI', i)
```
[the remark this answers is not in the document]
deliaBlue: I looked at the SAM file with multimappers and `XI` always start at 0. If wanting to start the hit index at 1, every entry would need to be changed.

scripts/filter_multimappers.py
deliaBlue: Won't it be better that instead of creating a new function I just modify the `write_output` one ? Looking at the code closer, the `write_output` function has the same conditions as this repeated part. The the function would look something like this:

```
def write_output(alignments: List[pysam.AlignedSegment]) -> None:
    """Write the output to the standard output (stdout).

    Args:
        alignments: alignments with the same query name
    """
    if len(alignments) == 1:
        sys.stdout.write(alignments[0].to_string() + '\n')
    else:
        best_alignments = find_best_alignments(alignments=alignments)
        for alignment in best_alignments:
            sys.stdout.write(alignment.to_string() + '\n')
```

Then the main function would look like this:

```
def main(sam_file: Path) -> None:
    """Filter multimappers by indels count.

    Args:
        sam_file: Path to the input SAM file.
    """
    with pysam.AlignmentFile(sam_file, "r") as samfile:

        sys.stdout.write(str(samfile.header))

        current_query = None
        current_alignments: list[pysam.AlignedSegment] = []
        
        for alignment in samfile:
 
            if alignment.is_secondary or alignment.is_supplementary:
                continue

            if current_query is None:
                current_query = alignment.query_name

            if current_query != alignment.query_name:
                write_output(alignments=best_alignments)
                current_query = alignment.query_name
                current_alignments = [alignment]
            else:
                current_alignments.append(alignment)

        write_output(alignments=current_alignments)
```

uniqueg: Yes, fine

uniqueg (approved):

PR #47, opened 2023-04-06
feat: temporarily disable isomiR counting

deliaBlue:


uniqueg (approved):
LGTM. Guess the.question is going to be in which order to merge. I'm happy to start with this one. Please let me know :)

deliaBlue:
Yeah, this seems a good one to start as now testing any change locally will take much less time and will create way less files :))

PR #48, opened 2023-04-06
test: clean up and simplify tests

deliaBlue:


uniqueg:
@deliaBlue: can you resolve the conflict that arose from merging #49?

uniqueg (changes requested):
LGTM, just the conflict

deliaBlue:
The deletion of the isomiRs rules led to the workflow not keeping the files `chr_size.txt` and `mirna_mature_filtered.bed`. They will be need most probably in others PR (at least `chr_size.txt` for sure on the extension one, not quite sure about the annotation file). If it is ok with you, I'll keep the rules but take them out of the `expected_output.md5`. If in the future we realize they are not needed, they will be removed from the workflow too.

uniqueg:
> The deletion of the isomiRs rules led to the workflow not keeping the files `chr_size.txt` and `mirna_mature_filtered.bed`. They will be need most probably in others PR (at least `chr_size.txt` for sure on the extension one, not quite sure about the annotation file). If it is ok with you, I'll keep the rules but take them out of the `expected_output.md5`. If in the future we realize they are not needed, they will be removed from the workflow too.

Sure

uniqueg (approved):

PR #50, opened 2023-04-09
build: update segemehl

### LINKED ISSUE #42: Update Segemehl to 0.3.4
The newest version of `Segemehl` includes both the aligned and unaligned sequences on the output file. In order to update `segemehl` version, a new rule removing these unaligned sequences must be added.

deliaBlue:
Closes #42


deliaBlue:
The expected output had to be updated as now all `M`s are `=`.
Also, rules for changing the `M`s in the `oligomap` output were added. Finally, rules to deleted the unmapped sequences were also added.

uniqueg (changes requested):
The PR adds 4 rules, with 2 and 2 of these running in parallel, at best. This means that each run of the workflow will take considerably longer, with the total time equaling the sum of how long it takes to remove the unmapped reads and modify the CIGAR strings.

Did you check the options that would allow us to avoid adding these extra rules? For example, `segemehl` 0.3.4 has an option `--unmatched <filename>` that allows us to write out unmapped reads to a different file (which we can then just ignore). For the CIGAR strings, you could try to identify which of the downstream scripts rely on the `=` characters instead of the `M` characters and, if possible, update these. If not (easily) possible, you could still try to see if instead of having two separate rules for transcriptome and genome mappings, you could run the rule instead on the merged alignments (before uncollapsing).

deliaBlue:
Regarding the option `--unmatched <filename>`,  I already try it out and, if I do recall well, it does not delete the unmatched alignments but instead, it creates an extra file with the unmapped sequences while keeping the unmapped sequences on the final alignment file.

As for the change between `M` and `=`, I think that it could be done after concatenating the genome and the transcriptome alignments. A thing worth noting that I did not see when doing the change is that in the CIGAR string, the `M` is either a match or a mismatch. This implies this change to be made in more than one command line; as I see it, a proper script is needed.

Now, altogether,  I run the script without removing the unmapped sequences and without making the exchange and the workflow fails once the alignments have been concatenated. I propose a single script that removes the unmapped sequences as well as perform the exchange. This will imply a single file scanning.

uniqueg:
But what is it that fails? Couldn't we change the failing script?

Also, I guess we could remove unmapped alignments when we are parsing the file anyway, in one of the filter scripts. We could add an extra option for that. Ideally the first one (the Perl script).

deliaBlue:
In the case of the unmapped sequences, it fails when using the script `sam_remove_duplicates_inferior_alignments_multimappers.1_6.pl` because the unmapped sequences do not have the `NM` tag. On solution is to add a line that removes the sequence is there is no `NM` tag.

For the `M/=` part, the error arise in the `filtering_multimappers.py`. The whole error is:

```
Traceback (most recent call last):
  File "../scripts/filter_multimappers.py", line 164, in <module>
    main(sam_file=args.infile)
  File "../scripts/filter_multimappers.py", line 154, in main
    write_output(alignments=current_alignments)
  File "../scripts/filter_multimappers.py", line 124, in write_output
    best_alignments = find_best_alignments(alignments=alignments)
TypeError: find_best_alignments() got an unexpected keyword argument 'alignments'
```

But I cannot see why this error appears.... But when there are only `M` or `=` it does not appear... Anyways, it could be also fixed in the Pearl script I guess.

uniqueg:
1. I think you can remove these lines in the Perl script already. However, use the flag to tell if a read is unmapped rather than just the absence of `NM`. Check the [SAM/BAM specs](https://samtools.github.io/hts-specs/SAMv1.pdf), it's the `0x4` bit.

deliaBlue:
I've been trying to modify the `CIGAR` string by comparing it with the previous alignment if the query ID is the same then I realized that there are no mismatches in any of the `oligomap` alignments; could it be that `oligomap` does not allow mismatches? If so, could I just replace all `M`s with an `=`? I tried to find `oligomap` documentation but I could not find it.

uniqueg:
Well, the problem with your script is quite obvious given the error message:

Here's the signature of `find_best_alignments()` (full definition [here](https://github.com/zavolanlab/mirflowz/blob/9295360769f53b9600ccc756cb6347db0f5f2fed/scripts/filter_multimappers.py#L89-L112)):

```python
def find_best_alignments(alns: List[pysam.AlignedSegment]) -> List[pysam.AlignedSegment]:
```

The error message is accurate: There _is_ no keyword argument `alignments` defined for that function, and therefore _whenever_ the [function is called](https://github.com/zavolanlab/mirflowz/blob/9295360769f53b9600ccc756cb6347db0f5f2fed/scripts/filter_multimappers.py#L124), this error is expected.

The _real_ question then is: Why is the error not raised (i.e., [line 124](https://github.com/zavolanlab/mirflowz/blob/9295360769f53b9600ccc756cb6347db0f5f2fed/scripts/filter_multimappers.py#L124) is not called) in some situations? The only ways how this can happen if either:
1. There are no alignments in the input file
2. There are no multimappers in the input file
4. The multimappers in the input file are not correctly assembled into the list of multimappers for a given read (could have multiple root causes)

It should be trivial to rule out points 1. and 2. It should also not be difficult to add some debug messages to the code and feed the script different synthetic input files to understand what is happening. Just make sure you do full accounting of every alignment. At every step you need to still have the expected number of alignments given your test case.

This is a very good example of why it is important to have unit tests, even for small scripts. The CIGAR string likely isn't even the issue here (I'm pretty sure `pysam` knows well how to deal with both the older and newer CIGAR specifications), but the fact that the script hasn't been sufficiently tested.

Please open another issue where you provide comprehensive unit tests for every function of your code. The unit tests should cover all expected behaviors and cover every single line of code. Use `pytest` to write and run your unit tests, and make sure you add the unit test execution to the CI pipeline.

From now on, make sure to write unit tests for every script/code that you write!

And while it's not really your responsibility to provide unit tests for scripts you didn't write, could you please create one separate issue for every script, on the topic of providing comprehensive unit tests for them, and then create a [task list issue](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists) where you list those individual issues in a list with check boxes? This will allow us to easily follow the progress of unit test coverage.

deliaBlue:
The individual unit tests are also meant to be needed for the bash scripts? These scripts are mostly a single-line command.

deliaBlue:
Putting the unit tests within the `test` directory does not feel right and create a directory named `tests` might lead to confusions so I  will create a subdirectory on the `scripts` one containing the test (an the ones to come) and the possible file(s) needed. If you think it should be in any other way, I'll change it to that.

uniqueg:
The typical location is indeed `tests`. But I think here it does make sense to put it in `scripts/tests` for the reason you mentioned. Good thinking

scripts/sam_remove_duplicates_inferior_alignments_multimappers.1_7.pl
uniqueg, line 1: Can you please:
- Change the v1.6 script (rather than creating a new copy of it) so that I can see only the changes
- Rename it such that it does not include the version
- Remove any old copies of the script

With version control, I don't see any benefit of keeping multiple files. It's also annoying to have the version in the filename, because it requires changes on the workflow side whenever there is a change.

Finally, if the entire file shows as new (instead of just the lines you modified), it makes it very hard for me to review.

uniqueg (changes requested):

uniqueg:
Also, can you please provde a PR description that briefly (in one or few bullet points) describes the changes. Just referencing the issue is not enough, in general, but especially here. There was a whole discussion going into this PR, so it'd be good to summarize that for context.

deliaBlue:
When updating `segemehl` to 3.4 the following changes were need:
- update `sam_remove_duplicates_infeior_alignments_multimappers.pl` in order to skip the unmapped sequences. This change was need because the new `segemehl` version outputs both the aligned and unaligned sequences.
- add the `-e` option in the mapping in order to return the old `CIGAR` string version. This change is made in order to match the same style as the `oligomap` mapping.

uniqueg (approved):

PR #51, opened 2023-04-11
style: update workflow descriptor formatting

### LINKED ISSUE #46: format workflow descriptior files
Fix typos, definitions and  add snakefmt autoformatter. 

To do the former one:
* Add the snakefmt autoformatter (which I think leaves two empty lines between rules) to a new file environment.dev.yml
* Update the docs to indicate that developers (but not users) should update their environment with the packages listed in that file
* Activate/run snakefmt
* Use [Snakemake's linting functionality](https://snakemake.readthedocs.io/en/stable/snakefiles/best_practices.html); resolve any issues
    Add the linting call to the CI pipeline; checks should only pass if linting passes

deliaBlue:
Closes #46.


test/test_format.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
uniqueg, line 1: I think this file is just an accidental copy of `test_lint.sh`. It seems have the same contents and doesn't seem to be used anywhere. Please remove.

test/test_lint.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
uniqueg, line 1: I wrote this horrible Bash script a few years ago to have a quick test for our new CI pipeline. It wasn't meant to stick around. And now it keeps on coming back to haunt me 🙈😅

1. We don't need the script. We are just executing a single command. Why wrap it? It's just complicating things and makes it more difficult to read/understand what's happening. Just put the command directly in the CI workflow. GitHub Actions will clean after itself, so no issues with that. And all the other stuff is just boilerplate to support the wrapper script, not the test itself.
2. Even if a script _were_ useful here (e.g., because we would need to run multiple complex commands that would just blow up and make the CI workflow difficult to read), we should write a clean Python script that we could then run with a test runner like `pytest`. And that is written specifically for the use case (this script was written for a use case where multiple commands were executed and where an explicit, manual clean up was necessary).

.github/workflows/CI.yml
```diff
@@ -30,9 +30,15 @@ jobs:
...
       - name: Update mirflowz env with root packages
         run: mamba env update -n mirflowz -f environment.root.yml
+      
+      - name: Update mirflowz env with dev packages
+        run: mamba env update -n mirflowz -f environment.dev.yml
         
       - name: Run test for rule graph
         run: bash test/test_rule_graph.sh
       
+      - name: Run test for snakemake lint
```
uniqueg, line 40: "Run linting" or "Run snakemake lint" will do. Or if you want to be very explicit: "Check workflow descriptor files for lints"

uniqueg (changes requested):

uniqueg (approved):

PR #64, opened 2023-04-15
refactor: modify output directories

deliaBlue:
The first line has to be removed because it does nothing.
The second line, has been changed to match the output directory; before, it was creating a directory named `.mirna_annotations`.

### DIFF
--- a/workflow/rules/prepare.smk
+++ b/workflow/rules/prepare.smk
@@ -57,7 +57,6 @@ rule trim_genome_seq_id:
     output:
         genome=os.path.join(config["output_dir"], "genome.processed.fa"),
     params:
-        dir_out=lambda wildcards, output: output[0][:-4],
         cluster_log=os.path.join(
             config["cluster_log"],
             "genome_process.log",
@@ -297,7 +296,7 @@ rule gfftobed:
         bed=os.path.join(config["output_dir"], "mirna_annotations.bed"),
     params:
         cluster_log=os.path.join(config["cluster_log"], "gfftobed.log"),
-        out_dir=lambda wildcards, input: input[0][:-4],
+        out_dir=lambda wildcards, input: input[0][:-21],
     log:
         os.path.join(config["local_log"], "gfftobed.log"),
     container:
--- a/workflow/rules/prepare.smk
+++ b/workflow/rules/prepare.smk
@@ -296,7 +296,7 @@ rule gfftobed:
         bed=os.path.join(config["output_dir"], "mirna_annotations.bed"),
     params:
         cluster_log=os.path.join(config["cluster_log"], "gfftobed.log"),
-        out_dir=lambda wildcards, input: input[0][:-21],
+        out_dir=lambda wildcards, input: input[0][:-22],
     log:
         os.path.join(config["local_log"], "gfftobed.log"),
     container:
--- a/workflow/rules/prepare.smk
+++ b/workflow/rules/prepare.smk
@@ -9,6 +9,8 @@
 
 import os
 
+from pathlib import Path
+
 ###############################################################################
 ### Global configuration
 ###############################################################################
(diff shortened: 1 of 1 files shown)


workflow/rules/prepare.smk
```diff
@@ -57,7 +57,6 @@ rule trim_genome_seq_id:
     output:
         genome=os.path.join(config["output_dir"], "genome.processed.fa"),
     params:
-        dir_out=lambda wildcards, output: output[0][:-4],
```
uniqueg, line 60: This is such weird code. Good riddance!

workflow/rules/prepare.smk
```diff
@@ -297,7 +296,7 @@ rule gfftobed:
         bed=os.path.join(config["output_dir"], "mirna_annotations.bed"),
     params:
         cluster_log=os.path.join(config["cluster_log"], "gfftobed.log"),
-        out_dir=lambda wildcards, input: input[0][:-4],
+        out_dir=lambda wildcards, input: input[0][:-22],
```
uniqueg, line 299: Well, good to get rid of this weird trimming of an exact amount of characters from a path that is a _parameter_ (and therefore, I guess, could be _anything_).

So let's not make this even worse by trimming an even larger amount of characters. This is really asking for trouble. What is it that you wanna do? Sure you shouldn't use something that `pathlib` could do perfectly well for you?

deliaBlue: When I ran `snakemake --lint`  this was the way to define directories as variables, as lambda functions; according to snakemake, this makes the workflow reproducible as the user can change the directory's name and it will still be stored where it must. If you prefer it, I can try to use `config["output_dir"]` and see if `snakemake --lint` does not rise any error.

uniqueg: Can you please tell me what value for `out_dir` you are expecting that this gives a sensible directory? And why you can be sure that we would always get a sensible directory here for any value of `out_dir`? Or, alternatively, how the value of `out_dir` is inaccessible for the user to set so that we could be absolutely _sure_ that it's always the value we expect here?

As for using lambdas: The problem I have is not with the use of lambdas, but with what the lambda function as implemented here is doing. If you want to derive a path from the value of `out_dir` (that is different from `out_dir`), then there should be better/safer ways of doing that, likely involving something like `pathlib` or similar. Of course that could still go into the lambda's body.

I think this discussion would be a lot clearer if you were to tell me what exactly you are trying to do here? Do you want to find out the parent directory of `out_dir` or something like that?

deliaBlue: The desired value for `out_dir` is the value stored in the config file as `output_dir`. Provided that the input file name won't change and will be `mirna_annotations.gff3` the indexing  selects the whole path for the input file until the beginning of the input file name _pre se_. 

This line is just extracting the set `output_dir` using the whole path of the input file. I used this line as it was the one in the snakemake documentation.

uniqueg: Okay, maybe it's not likely that a user will accidentally change this name. However, if they were, or we were, during development, it could lead to bugs that are hard to trace back.

If I understand you correctly, what you want to have (and which I suspected) is what is typically known as the `dirname` (or at least that's the name of the Bash/GNU program that gives you that), i.e., the directory that contains your file.

So instead of having this weird code with the `22` hardcoded, you can use `Pathlib` and get the `.parent`, like so:

```python
from pathlib import Path

out_dir=lambda wildcards, input: str(Path(input[0]).parent)
```

This will give you the name of the directory that contains your file. If you need to go up more directories, you can use `.parents[i]`, where `i` gives you the `i`th directory up, starting with `0` as usual (`.parent` is just a shortcut for `.parents[0]`). Always use `pathlib` when dealing with paths, it's much better than `os.path` (although that's still a lot better than manipulating strings directly).

Possibly the `str()` is not needed - depends on whether Snakemake knows how to deal with `Path` objects. At this point, I suspect, it probably does.

uniqueg (approved):

PR #72, opened 2023-04-19
test: add unit tests

deliaBlue:
Closes #52 

In order to add a test for `filter_multimappers.py` the following changes had been made:
- Create the test
- Create the test files (input and output files)
- modify the script to stop if the file is empty and fix `count_indels` and `find_best_alignments` calls
- Add dependencies in both the `environment.yml` and `environment.dev.yml`
- Add test in `CI.yml`
- Add coverage file

Issue #69, opened 2023-04-17
Intersect BAM with extended miRs.


deliaBlue:
As in issue #66 , the first rule will use `bedtools intersect` with the options  `-wb` and `-F 1` but we will be using the extended miRs annotations. 

The second rule, will use the 11th column to add to the previous SAM file the name of the mature miRNA(s) an alignment overlaps the most; the name will be added as a new tag.

deliaBlue:
This issue must also include the conversion of the BAM file into a BED file as the first step.

uniqueg:
Can't `bedtools` handle BAM or at least SAM files?

deliaBlue:
Closed in PR #84 .

PR #73, opened 2023-04-27
feat: extend mature miRNA annotations

### LINKED ISSUE #65: Extend miRNA n nucleotides.
Extend miRNAs by `n` nucleotides using `gffutils`. 
`n` will be a parameter for the user to define which default (and possible max value) will be determined by setting different values, running the workflow and analyzing the results for several libraries; if no conclusion can be taken from there, the default will be set to 6 (while finishing the whole workflow, this will be the default value used). 

pre-miRs will be extended if necessary; that is, if the canonical miRNA exceeds the start/end of the pre-miRs. Moreover, the names of the pre-miRs will be updated to show its (possible) extension. 

This new rule - which will substitute the  extract mature mirs - will produce two different output files; one for the mature miRNA and one for the pre-miRs annotations.

deliaBlue:
Closes #65 

In this PR the changes applied are:
- Creation of a class that extends mature miRNA coordinates
- Implement a script that uses the class
- Create unit tests for both the class and the script
- Update dev dependencies to include `gffutils` and the `pytest` and `coverage`


deliaBlue:
If looking at the unit tests report, there are two things that I do not quite understand:
- First of all, it says that not all lines of code are being tested, but they are; in specific it says that all lines in the `mirnaExtension_class.py` that do all the work when mature miRNAs are found.
- Then, there are a bunch of warnings; should  I ignore them? I do not understand why they are appearing but it seems like I can "capture" so they can be ignored (?).

```

=============================== warnings summary ===============================
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:121
  /usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:121: DeprecationWarning: pkg_resources is deprecated as an API
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)

../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
  /usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
  /usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google.cloud')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2349
../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2349
  /usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2349: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(parent)

../../../../../../../usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870
  /usr/share/miniconda/envs/mirflowz/lib/python3.9/site-packages/pkg_resources/__init__.py:2870: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google.logging')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

deliaBlue:
The expected output was changed because the output of the primary transcript annotation file changed with the refactoring to change the primary transcript names if there has been an extension.

uniqueg (changes requested):
This PR should not include changes in 21 files. Can you please either rebase this on `dev` and add only what is needed. Alternatively, we will need to wait until #72 is merged and _then_ rebase/merge with that.

deliaBlue:
I think I'll wait for the merge with #72. There are a lot of files, but mind that the testing files alone are more than 10 as each input file requires two different output files. Still, I'll wait for that merge and see what needs to be changed.

deliaBlue:
I know there are a lot of files changed but all the changes are needed:
- 11 files are for the testing + the test file
- 2 files are the new scripts
- 3 Snakemake files had to be modified to integrate the new scripts and their outputs + the expected output needed to change too
- the config file had to change so it allow the workflow to run without errors
- the environment.dev.yml needed to have gffutils
- finally, the CI "changed" when doing the merging but you'll see it's just whitespaces

As a final note, by the moment, I did not put the selection of the value `n` in the config file but directly as a parameter in the script call; I can change that before you take a look at the changes already in the PR or I can do it after. It makes more sense to do it before but given that you see there are too many changes already, maybe you prefer to apply this last one after the first review.

scripts/mirna_extension.py
```diff
@@ -0,0 +1,124 @@
...
+
+    m = MirnaExtension(
+        gff_file=args.input,
+        premir_out=args.premir,
+        mir_out=args.mir,
+        n=args.extension,
+        seq_lengths=seq_lengths
+        )
+    m.load_gff_file()
+    m.extend_mirnas()
```
uniqueg, line 118: Not all variables used in a class should be (instance) attributes. Attributes are for passing data around between methods. I think if you have a method `.load_gff_file()`, then this is basically a setter method for the `db` property and it makes more sense to me to pass the file path as an argument to that method because it is likely not used elsewhere. Likewise, the extension paramter is probably only used in `.extend_mirnas()` and so it does not make sense to put it as an (instance) attribute. I mean, why would I need to create a new instance of the class to create miRNA extensions with a _different_ extension? About the other attributes I'm not sure. But generally, if you use them in only one method, then define them as args there.

test/config.yaml
```diff
@@ -3,4 +3,4 @@ genome_file: test_files/genome.fa.gz
 gtf_file: test_files/gene_annotations.gtf.gz
 mirna_file: test_files/mirna_annotations.gff3
 map_chr_file: test_files/ucsc_to_ensembl.txt
-mir_list: ["miRNA", "miRNA_primary_transcript"]
\ No newline at end of file
+mir_list: ["miRNA"]
```
uniqueg, line 6: Why does this not work anymore with `"miRNA_primary_transcript"`?

uniqueg (changes requested):
I'll review more after you've addressed the comments

test/config.yaml
deliaBlue: This does not work anymore with "miRNA_primary_transcript" as now the annotation file does only contain either "miRNA" or "miRNA_primary_transcript" and when merging the tables it looks for them. I thought that for this PR I would delete the "miRNA_primary_transcript" and two issues ahead (#69) when extending the BAM file with the extended primiRs, I would add it again and uncomment the pri-mirs table rule.

scripts/mirna_extension.py
```diff
@@ -0,0 +1,124 @@
...
+        help="Path to the gff3 pre-miR output file",
+        type=str,
+        required=True
+    )
+    parser.add_argument(
+        '--mir',
+        help="Path to the gff3 miRNA output file.",
+        type=str,
+        required=True
+    )
```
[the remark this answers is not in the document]
deliaBlue: right now, we are using as an input file one called `mirna_annotations.gff3` wouldn't be weird to have as a file name `mirna_annotations_mirs_extended_6_nt_mature.gff3`? wouldn't be best if we set the name to directly to `mirna_annotations_extended_6_nt_mature.gff3` independently of the input file?

test/config.yaml
uniqueg: Couldn't you just add the new script/rule without changing/removing any other rules? Removing something isn't really part of this issue. We can clean up later.

uniqueg (changes requested):

try/collapse_sam_by_seq.py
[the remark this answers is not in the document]
deliaBlue: Sorry about that, I don't know how this end up here

test/expected_output.md5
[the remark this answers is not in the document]
deliaBlue: I don't know why the new files do not appear in the final files, even when added to the final rule of both the main snakefile and the `prepare.smk`; maybe when using them later on as input files they might appear.

scripts/mirna_extension.py
```diff
@@ -57,14 +54,14 @@ def load_gff_file(self, gff_file: str = "") -> None:
                 path to the input GFF3 file. If None, input is taken from the
                 standard input (stdin).
         """
-        if gff_file == "":
+        if gff_file == None:
```
uniqueg, line 57: Test against `None` is with `is` not `==`

scripts/mirna_extension.py
```diff
@@ -189,17 +186,16 @@ def main(args):
...
         with open(args.chr, 'r') as f:
             for line in f:
                 ref_seq, length = line.strip().split("\t")
                 seq_lengths[ref_seq] = int(length)
-    else:
-        seq_lengths = None
 
     m = MirnaExtension()
-    m.load_gff_file(args.input)
+    m.load_gff_file(str(args.input))
```
uniqueg, line 198: I'd use type `Optional[Path]` in `load_gff_file()` as well. If `gffutils.create_db()` cannot deal with `pathlib.Path`, then you can still convert with `str()`.

scripts/mirna_extension.py
```diff
@@ -23,15 +23,12 @@
 import os
```
uniqueg, line 23: Do you still need this? If so, see if you can use `pathlib.Path` instead. No point in importing a rather large module if we don't really need it.

uniqueg (changes requested):
Getting there. 3 more minor comments, I think after we can merge :)

deliaBlue:
If I am not wrong, we said that the extension would be for the user to define; provide is option on the config file has not yet been implemented. Do you still want this option to be for the user to change or you rather leave it at 6 by the moment?

uniqueg:
> If I am not wrong, we said that the extension would be for the user to define; provide this option on the config file has not yet been implemented. Do you still want the extension to be for the user to change or you rather leave it at 6 by the moment?

Yes, it should indeed by up for the user, good that you remembered. Please add it to the config file; this PR is the right place to do this.

deliaBlue:
While adding the `extension` option I saw that ''config_schema.json` and `config_template.yaml` are outdated. They need to have removed the "isomiRs generation parameters" and in case of the template, it need to have the addition of the `extension` parameter. Can I do a simple PR named "build: update config files" and do these two changes?

uniqueg (approved):

PR #75, opened 2023-05-02
refactor: intersect alignments with extended premiRs

### LINKED ISSUE #66: Intersect BAM with extended pre-miRs.
This step will be done in two different rules. 

The first rule - which will substitute the `intersect_mirna rule` - will use `bedtools intersect` with the options `-wb` to keep the full records of `-b` and `-F` will be set as 1 to make sure that the whole alignment is within the (extended) pre-miRs regions. 

The second rule will use the 14th column of the intersection output  file to create a new SAM file containing only the intersecting alignments.

deliaBlue:
Closes #66 .

In order to merge the alignments with the extended premiR annotation files the following changes had been made:
- Turn the extended premiR annotation GFF3 file into a BED file
- Intersect the alignment files with the created BED file
- Use the output file to create the premirs table
- Filter the SAM file alignments to only contain the ones that intersect  (needed in future steps)
- Comment the rule creating the miRNAs table
- Remove the miRNAs from the config file so it does not raise an error when merging tables
- Add the new files in the main Snakefile
- Update the expected output


uniqueg (changes requested):
Can you please update the rule graph so that I can see visually what is happening? Please remember to do that whenever you change the workflow.

Also, it seems to me that since this PR is changing the wiring only, you might as well use that to address whichever issues you found with the config schema and template - it's a good enough fit for me to update some old leftover things that you found in the wiring.

uniqueg (approved):

PR #82, opened 2023-05-16
refactor: use collapsed reads throughout workflow

### LINKED ISSUE #81: Use collapsed SAM file in bedtools intersect.
In this issue, changes in the wiring and in the pre-miR counting steps will be assessed.

deliaBlue:
Closes #81.

This PR has the following changes:
- Usage of the collapsed files for the entire workflow
- Uncollapse final SAM file and conversion into a BAM one
- Modify the `primir_quantification.py` script with the avoidance of the usage of the SAM file and the `--collpased` option addition
- Modify the unit tests for the `primir_quantification.py`
- Update of the unit tests files (in/out files + removing the unused ones)
- Update the rule graph to depict the new workflow
- Update the expected output to contain the new files


scripts/primir_quantification.py
```diff
@@ -45,9 +42,11 @@ def parse_arguments():
         type=Path
     )
     parser.add_argument(
-        '-s', '--sam',
-        help="Path to the SAM file containing the intersecting alignments.",
-        type=Path
+        '--collapsed',
+        help="Indicate that the file used in bedtools intersect has the\
+            reads collapsed by sequence and alignment. Default %(default)s.",
```
uniqueg, line 47: It's better to use parentheses rather than `\` for splitting lines. For example:

```python
help=(
    "Line 1 continues"
    " on line 2."
)
```

Also, you need to indicate what's the expected format/syntax for collapsed reads (as produced by the FASTX collapser).

scripts/primir_quantification.py
```diff
@@ -125,13 +99,10 @@ def main(args) -> None:
             line = line.strip().split('\t')
             name = attributes_dictionary(line[9])[args.id]
```
uniqueg, line 100: Using these indices always makes me a bit uneasy. It makes the script hard to read/maintain/review. And we've had one nasty bug once in one of our pipelines because of a wrong index. Both columns ended up being numbers, but the wrong number was used. And as they were not totally different in scale, it took a long time until we found that mistake, and everything that we ran through that script was wrong :)

Of course, these kinds of problems can also occur with named columns, but they are less likely, because you can introduce such a problem only when the column names are assigned, as opposed to everytime a value is called. Also, because you need to define a tuple of column names, and it has to match the length of the list, it is not really possible to skip or include an extra column. And finally, when defining column names, it is more likely that one pays extra attention not to mess up the order.

If you only need to read from the list of values and not write back to it, the best data structure to use is probably a `namedtuple`. Otherwise a dictionary will of course work, which you can `zip()` together from your tuple of names and the row data.

Alternatively, you could use NumPy arrays or Pandas series. These may be more convenient later on, depending what you are doing with the data.

uniqueg (changes requested):
Excellent PR! :)

Just some minor issues

scripts/primir_quantification.py
deliaBlue: There are some columns on the output bed file that I do not know what they are. I've been looking on several places and some of them came clear but others don't, right now the naming looks like this:

```
fileds = namedtuple("feat_chr", "feat_start", "feat_end", "feat_id", "score", "strand", "?", "feat_type", "?", "feat_attributes", "read_chr", "read_start", "read_end", "read_name", "NH", "read_strand",  "intersection_start", "intersection_end", "itemRgb", "intersection_fraction", "intersection_length", "?")
```

uniqueg: Can you post a few lines of an example file and the corresponding `bedtools intersect` call that produced that file?

deliaBlue: The `bedtools intersect` call is:

```
bedtools intersect  -wb  -s  -F 1  -sorted  -b BED12  -a BED  > output.bed
```
Some lines of the output file are:

```
19	88	108	MI0003140	.	+	.	miRNA_primary_transcript	.	ID=MI0003140;Alias=MI0003140;Name=hsa-mir-512-1_-0_+0	19	88	108	766380-1	3	+	88	108	255,0,0	1	20	0
19	88	109	MI0003140	.	+	.	miRNA_primary_transcript	.	ID=MI0003140;Alias=MI0003140;Name=hsa-mir-512-1_-0_+0	19	88	109	535182-1	3	+	88	109	255,0,0	1	21	0
19	54439	54462	MI0003165	.	+	.	miRNA_primary_transcript	.	ID=MI0003165;Alias=MI0003165;Name=hsa-mir-517b_-1_+3	19	54439	54462	270359-1	5	+	54439	54462	255,0,0	1	23	0
```

deliaBlue:
The new changes in the PR are related with using `BAM` file in the `bedtools intersect` call. The changes are:

- Add the `NH` tag value on the alignments names in the `filter_multimappers.py` script
- Delete the `BED12` file creation
- Use the `BAM` file in `bedtools intersect` call
- Add the `--nh` option in the `primir_quantification.py` script
- Update the unit test for both scripts as well as the test files
- Update the expected output
- Update the rule graph

uniqueg (approved):

Issue #71, opened 2023-04-17
Create canonical/isomiRs table


deliaBlue:
The output of issue #69 along with `pysam` will be used to create a tabulated file with the following fields:

- Name of miR species decied in issue #27  if an isomiR, otherwise, the canonical name (str)
- Product of the amount of contributed reads and  1 divided by the number of best alignments (float)
- Number of best alignments being the `NH` tag (int)
- Name of every contributing read (str)
- Length of the read (int)

PR #88, opened 2023-05-24
feat: (iso)miR quantification

deliaBlue:
This PR adds a boundary for the 5p-shift and the 3p-shift between a feature and an alignment to be added to the name. In addition, the 3p-shift calculation has been fixed along with some style changes.


deliaBlue:
The new modifications in this PR closes #71 by applying the following changes:
- Add rule to order the final `SAM` file by the tag containing the intersecting miRNAs in `quantify.smk`
- Add script `iso_mirna_quantification.py` to create (iso)miR table
- Add rule to create the (iso)miR table in `quantify.smk`
- Delete previous quantification script `mirna_quantification.py`
- Add unit test script for `iso_mirna_quantification.py`
- Add unit test files
- Change output files in `Snakefie`
- Add `miRNA` in the `config.yaml` as a feature to quantificate
- Update expected output

deliaBlue:
The last changes in this PR allow the user to define which type of miRNA wants to have and in which format. The modifications are:
- Change the names in the `config.yaml` for the miRNA type
- Add two functions in `iso_mirna_quantification.py` to create the selected directories and to write to them
- Modify the `quant_mirna` rule in `quantify.smk` to allow the type selection
- Add unit test for this new functionality
- Add test files
-  Modify the `config_template.yaml` and `config_schema.json`
-  Update the expected output

scripts/iso_mirna_quantification.py
```diff
@@ -0,0 +1,337 @@
...
+            "Default %(default)s."
+        ),
+        default='YW',
+        type=str
+    )
+    parser.add_argument(
+        '-c', '--collapsed',
+        help=(
+            "Indicate that the SAM file has the reads collapsed by sequence"
+            "and alignment. The collapsed name must be build by the alignment"
```
uniqueg, line 100: By sequence and alignment? What do you mean? I thought it's just the reads collapsed with `fastx-collapser`. You can reference that also here, e.g.:

> ", as, e.g., produced by the 'fastx_collapser' tool of the FASTX-Toolkit: http://hannonlab.cshl.edu/fastx_toolkit/".

To give you one example here for a thorough description (but please reconsider the descriptions for _all_ CLI options):

"Set to enable quantification if reads in SAM file were collapsed. In that case, SAM query names are expected to follow the general format 'NAME-COUNT', where NAME is an arbitrary unique name and COUNT is the number of identical reads that were collapsed. For example, the query name 'my_query_name-123' would indicate that the read represented by this alignment occurred 123 times before collapsing. The required naming format is produced, e.g., by the 'fastx_collapser' tool of the FASTX-Toolkit: http://hannonlab.cshl.edu/fastx_toolkit/"

scripts/iso_mirna_quantification.py
```diff
@@ -0,0 +1,337 @@
...
+            "The name must be build by the alignment name followed by an"
+            "underscore and the NH value, i.e 1-2_4. Default %(default)s."
+        ),
+        action='store_true',
+        default=False
+    )
+    parser.add_argument(
+        '--len',
+        help=(
+            "Include the species length in the output table."
```
uniqueg, line 120: What species?

scripts/iso_mirna_quantification.py
```diff
@@ -0,0 +1,337 @@
...
+    iso_mirna_out = None
+
+    if 'mirna' in mir_list:
+        mir_out = outdir/f'mirna_counts_{library}'
+    if 'isomir' in mir_list:
+        isomir_out = outdir/f'isomir_counts_{library}'
+    if 'iso_mirna' in mir_list:
+        iso_mirna_out = outdir/f'iso_mirna_counts_{library}'
+
+    return mir_out, isomir_out, iso_mirna_out
```
uniqueg, line 167: Why make this so complicated? There's only _one_ output file, that either does or does not contain canonical and/or isomiRs. The user will get what they ask for and see for themselves which feautures are quantified in the output. No need to indicate that in the filename. Just call it `counts_mir_{library}.tsv` )versus `counts_pri-mir_{library}.tsv` for the primary miRNA quantification). When starting your names with the more general term (here: `counts`), in the final workflow output directory, the count tables appear right next to each other.

scripts/iso_mirna_quantification.py
```diff
@@ -0,0 +1,337 @@
...
+        nh:
+            indicates if the NH vale is found on the alignment name
+    Returns:
+        the contribution of the alignment to the overall total
+    """
+    query_id = aln.query_name
+
+    if collapsed and nh:
+        num_reads = int(query_id.split('-')[1].split('_')[0])
+        nh_value = int(query_id.split('-')[1].split('_')[1])
```
uniqueg, line 196: This will fail for many query names, e.g., for `my_funny-query_name-123_4`. Use a regular expression instead. You will need to have `_NH` pinned at the end and `-COUNT` either at the end _or_ followed by `_NH`.

scripts/iso_mirna_quantification.py
```diff
@@ -0,0 +1,337 @@
...
+            nh_value = 1
+
+    else:
+        num_reads = 1
+        try:
+            nh_value = aln.get_tag("NH")
+        except KeyError:
+            nh_value = 1
+
+    return num_reads/nh_value
```
uniqueg, line 216: This algorithm will be super slow for cases where neither NH nor the number of collapsed reads are used. It will actually be the slowest of all possible modes, even though it _should_ be the fastest (because, if the options are _not_ indicated, we should never need to check that for any read).

Best is if you can find a way to avoid all these conditionals completely in the loop. You could generate a function that always returns you the values for NH and the number of counts either given a regex or a default, then map that function as an input to a general counting function that simply does `num_reads/nh_value` to your alignments.

scripts/iso_mirna_quantification.py
```diff
@@ -0,0 +1,337 @@
...
+    Args:
+        pre_name:
+            string with the format feat_name|5p-shift|3p-shift|CIGAR|MD
+
+    Returns:
+        list with the species name to be found in the final table and its type
+    """
+    data_name = pre_name.split("|")
+
+    if data_name[1] == '0' and data_name[2] == '0' and data_name[3][:2] == data_name[4]:
```
uniqueg, line 237: I think the CIGAR and MD check is really lazy. You should check them fully and individually, then compare the number for equality. And don't rely on their length to be always double digit. Even if it's going to be always true (which you never know), I think it's inelegant to hardcode such assumptions.

uniqueg (changes requested):

config/config_schema.json
```diff
@@ -92,7 +92,7 @@
         },
         "mir_list":{
             "type": "array",
-            "default": ["miRNA", "miRNA_primary_transcript", "isomirs"]
+            "default": ["iso_mirna", "pri_mir"]
```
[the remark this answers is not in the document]
deliaBlue: I have to use `mirna` instead of `mir` because otherwise, the script `merge_tables.R` takes the `iso` part of `isomir` as part of the library name when merging the `mir` tables  (idk why)

scripts/iso_name_tagging.py
```diff
@@ -61,7 +61,7 @@ def parse_arguments():
     parser.add_argument(
         '--id',
         help=(
-            "ID used to identify the feature in the output table."
+            "ID used to identify the feature in the name add as tag."
```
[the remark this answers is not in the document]
deliaBlue: Maybe if I change "add" by "added"? What I am trying to say is that the selected `--id` will be the one used to name the intersecting feature that will, later on in the script, be add to the alignment as a tag.

scripts/iso_mirna_quantification.py
uniqueg: Basically, there are roughly two approaches you can follow:

1. You define 4 different counting/contribution functions, one each for the combined states of `collapsed` and `nh` (which either can be `True` or `False`). Then, _before_ you loop over the individual alignments, you determine which of the four functions is the function you need to use here. Then when looping over the alignments, you apply that function to your read.
2. Similar approach but you only define one very simple counting/contribution function that takes values for number of collapsed reads and the number of hits inputs and returns their quotient, i.e., $n_{reads}/n_{hits}$. And then you have to populate the input params with an appropriate regex. But as you would need to fall back to defaults if NH or number of reads is not part of the query name, you'd probably end up having to write small functions for each case here as well, so the benefit over 1 may be minimal. Besides, in three of the four cases defaults (i.e., 1) are used for either or both of the two params, so always calculating the quotient is not necessary or at least will carry the overhead of defining a variable where none is needed.

uniqueg (changes requested):

images/rule_graph.svg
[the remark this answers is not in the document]
deliaBlue: Since this changes do not affect the workflow _per se_, I'll open an issue to fix these output files/rule naming and another one to add the workflow documentation (or maybe both things in a single one?) and I can do it in parallel to the other things if that's alright.

uniqueg (changes requested):
I couldn't review it _super_ deeply, but I didn't see any obvious issues, at least nothing major. I'll have another look tomorrow.

workflow/rules/quantify.smk
```diff
@@ -468,35 +496,35 @@ rule quant_mirna_pri:
...
-       --prefix {params.prefix} \
-       --verbose \
-       ) &> {log}"
+    input:
+        table=expand(
+            os.path.join(
+                config["output_dir"], "TABLES", "{mir}_counts_{sample}"
+            ),
+            sample=pd.unique(samples_table.index.values),
+            mir=[mir for mir in config["mir_list"] if mir != 'isomir'],
```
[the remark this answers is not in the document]
deliaBlue: In order to avoid hardcoding it I tried using [specific wildcards constrains](https://snakemake.readthedocs.io/en/stable/tutorial/additional_features.html#constraining-wildcards) as well as the [general wildcards constrains](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#wildcards). None of them worked. I can either try to find another way now or maybe open an issue for this (?)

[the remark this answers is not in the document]
uniqueg: No, as I said, just leave it :)

scripts/iso_name_tagging.py
```diff
@@ -1,17 +1,30 @@
...
+must contian only the reads that have an intersecting feature.
+
+NAME CREATION and TAG ADDITION
+For each alignment, the name of the the intersecting feature will follow the
+format FEATURE_ID|5p-shift|3p-shift|CIGAR|MD. The CLI option `--id` specifies
+the feature identifier to be used as FEATURE_ID from within the attributes
+column in the BED file. The 5p-shift and the 3-p shift values are the
+difference between the feature start and end coordinates and the alignment
+start and end coordinates. If `--extension` is provided, the feature start
+position by are adjusted by adding the given value and subtracting it from the
```
uniqueg, line 24: Remove first "by"

scripts/iso_name_tagging.py
```diff
@@ -1,17 +1,30 @@
...
+end position. If both, the 5p-shift and the 3p-shift, are within the range 
++/- extension + 1 the feature name is added to the alignment as the new tag
+"YW". Multiple intersecting feature names are separated by a semi-colon.
```
uniqueg, line 27: You _absolutely_ need to give an example involving a BED record, a SAM record, the construted name and the new SAM record with the tag. And make sure the example covers all the special stuff, like extension etc. Or have multiple examples.

scripts/mirna_quantification.py
```diff
@@ -1,163 +1,455 @@
...
+    input: SAM meeting the minimal characteristics
+    command: mirna_quantification.py SAM --mir-list mirna
+    output: hsa-miR-517b-3p	1.3333333333333333
+            hsa-miR-518b	1.0
+
+    Example 3
+    input: SAM meeting the minimal characteristics
+    command: mirna_quantification.py SAM --mir-list mirna
+    output: hsa-miR-512-3p|0|1|23M|22C	1.6666666666666665
+            hsa-miR-517-5p|-1|0|23M|22T	0.6000000000000001
```
uniqueg, line 84: Sure about this?

uniqueg (approved):
I'm already approving because I'm sure you can do these minor things without another review. And you can add the option for the column, if you like :)

PR #96, opened 2023-07-02
build: use biocontainer `oligomap` image

deliaBlue:
In this PR the docker image for `oligomap` is changed from the zavolab container to the one in `bioContainers`.


uniqueg (approved):
😍

PR #111, opened 2023-09-19
build: create independent environment files

deliaBlue:
This PR makes the following changes:

- Create four independent environment files
- Update README with the correct installation steps
- Update CI to meet the changes
- Update the `gffutils` micro-environment version
- Add `nodefaults` channel in the environment files (to avoid the `defaults` channel included by GitHub)


uniqueg (changes requested):
Better implement/refactor as discussed on Slack

PR #114, opened 2023-10-25
ci: fix failing static code analysis tests

deliaBlue:
This PR has the following changes:

- Correct the return type in `iso_name_tagging.py`
- Modify some variables to be strings in `mirna_quantification.py`
- Upgrade and pin some dependencies version in `environment.dev.yml`


uniqueg (changes requested):
Why do you feel it necessary to pin _all_ dependencies_? This is asking for trouble. Every time any security vulnerability is found in any of these, we will need to update that dependency - which typically requires other dependencies to be updated as well, and it's a pain to figure out one specific configuration that works.I would rather exclude only those verions of individual dependencies that I know that cause problems. Or, at least, set ranges where you exclude the next major versions, e.g., `coverage >=7.2, <8`.

uniqueg (approved):

PR #115, opened 2023-10-30
refactor: oligomap output conversion script

### LINKED ISSUE #54: refactor: refactor and create unit test for `oligomapOutputToSam_nhfiltered.py`
Provide comprehensive unit tests for every function in the code by formalizing the generation of synthetic inputs and checking against expectation.
The unit tests must cover every single line of code.
Use `pytest` to write and run the unit tests, and use the `pytest-cov` extension to track the test coverage.
Add the unit test execution and coverage calculation to the CI pipeline.

deliaBlue:
This PR closes #54 .

The modifications in this PR are:

- Refactor and rename the script `oligomap_output_to_sam_nh_filtered.py` 
- Create the unit test for the script and add the necessary files
- Modify the CI file to include the script in the static code analysis
- Modify the subworkflow `map.smk` to accommodate the refactoring
- Modify the link to the `oligomap_output_to_sam_nh_filtered.py` in the `pipeline_documentation.md`


deliaBlue:
The static code analysis test fails as the PR #114 is not merged yet. Once it is, and merge into this PR, it will pass.

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: flake8
         working-directory: ./scripts
-        run: flake8 ./*.py --exclude oligomapOutputToSam_nhfiltered.py
+        run: flake8 ./*.py
```
uniqueg, line 41: Why include `./*.py` - by default `flake8` only checks `.py` files, no?

scripts/oligomap_output_to_sam_nh_filtered.py
uniqueg, line 1: Very hard to review when the changes are so extensive and the original file was deleted (renamed). There is no diff view.

scripts/oligomap_output_to_sam_nh_filtered.py
```diff
@@ -0,0 +1,450 @@
+#!/usr/bin/env python
+
+"""Transform oligomap output FASTA file to SAM keeping the best alignments.
+
+Read the input FASTA file, pick the best alignment(s) for each read (i.e. all
```
uniqueg, line 5: The `oligomap` output is a proprietory/custom file format. It is NOT FASTA.

scripts/oligomap_output_to_sam_nh_filtered.py
```diff
@@ -0,0 +1,450 @@
...
+the alignment number of errors and reference sequence strand.
+- The fourth and sixth lines must be the read and reference sequence
+respectively.
+- The fifth line must contain a vertical bar for each match between the
+sequences base pairs and a space otherwise.
+
+Alignments must be separated with a blank line.
+In addition, alignments have to be sorted by the read name and separated
+with a blank line. An example of how an entry would look like in the EXAMPLES
+section.
```
uniqueg, line 29: It would actually make more sense to provide this information on the `oligomap` project page rather than here.

scripts/oligomap_output_to_sam_nh_filtered.py
```diff
@@ -0,0 +1,450 @@
...
+The mapping quality, `map_q`, is set to 255. Both, the position of the next
+read, `p_next`. and the observed template length,`t_len`, are set to 0.
+Both, the reference name of the next read, `r_next`, and the ASCII of
+Phred-scaled base quality, `qual`, are set to '*'. In addition, the
+optional fields, edit distance, MD string, and number of hits are also
+included. Only the best alignments per read (i.e. either all the alignments
+have 0 or 1 error) are written to the output file. Moreover, if the
+`--nh-filter` CLI argument is given, reads with more htis than the provided
+value are discarded. If none of the alignments meet the criteria, an empty
+file is created.
```
uniqueg, line 45: Would be better to have this in table form. It's difficult to parse otherwise

workflow/rules/map.smk
```diff
@@ -518,9 +517,11 @@ rule convert_genome_to_sam_oligomap:
         os.path.join(workflow.basedir, "envs", "python.yaml")
     shell:
         "(python {input.script} \
-        -i {input.sort} \
+        {input.sort} \
         -n {params.nh} \
-        > {output.gmap}) &> {log}"
+        --ref {params.ref} \
+        --outdir {params.out_dir} \
+        ) &> {log}"
```
uniqueg, line 524: Why did you change the API of the script? Why is `--ref` needed now but wasn't before? Not saying it's a bad thing, but you should probably include the reasoning for this in the PR description.

More problematic: Why do we have an output directory now, instead of just a single file? Are you creating more than one output file now? If so, why?

workflow/rules/map.smk
```diff
@@ -505,6 +502,8 @@ rule convert_genome_to_sam_oligomap:
             config["cluster_log"], "oligomap_genome_to_sam_{sample}.log"
         ),
         nh=config["nh"],
+        out_dir=lambda wildcards, output: Path(output[0]).parent,
```
uniqueg, line 505: I find this difficult to parse and accessing the `0` index possibly dangerous. What exactly happens here? And how does this connect to `output.gmap`? See comment below also (not clear to me why we don't still have just a single output file, much cleaner)

scripts/oligomap_output_to_sam_nh_filtered.py
uniqueg: Also, this really looks like quite an extensive rewrite. That's not necessarily bad. But I'm wondering: How do you know the new script does what the old script was supposed to do when you didn't have unit tests set up prior to refactoring?

uniqueg (changes requested):
I'll leave it at this for now. Let's discuss the comments I made so far before I dig deeper into the code/changes.

scripts/oligomap_output_to_sam_nh_filtered.py
deliaBlue: I can try to restore the original script and use `git mv` and paste the contents of the refactored code if it is easier for the review. I changed the name locally and made the first commit without taking this into account. Sorry about that.

As for how I know it does the same, during the refactoring I followed these steps:

1. Identify the smaller functional units of code to create the functions and find the relations between them.
2. Identify what elements can be fit inside the `main()` function.
3. Make the modifications needed to fit the new structure without modifying the script _per se_ unless necessary
4. Use `flake8`, `mypy` and `pytest` to see what data types, variables and function arguments have to be modified. Apply the changes.

On steps 3 and 4, I used the output of two different test libraries provided by the old script with to compare the output of the refactored one. This helped me on seeing what changes made the script behave differently, referring to the old one to check the specific code snippet that was producing this difference, and fix it.

Maybe that is not the best procedure, but is the one I though it will keep the "essence" of the old script while refactoring it to the format of the new ones.

scripts/oligomap_output_to_sam_nh_filtered.py
deliaBlue: Oh! As in the workflow and the `oligomap` documentation the output is always redirected to a  FASTA, I thought that was the format to be set in the output file. Changing that

scripts/oligomap_output_to_sam_nh_filtered.py
deliaBlue: I'll put this information in the main README of the `oligomap` repository and redirect the user of this script there.

workflow/rules/map.smk
deliaBlue: If being totally honest, I remember that I made this change for a reason but cannot recall it right now, apologies...

Whatever reason was behind using the output directory, I had to use the `--ref` flag because in order to write to the output file, I needed to give it a name, and depending on the source of the mappings - either genome or transcriptome - the file name changes. Should I simply write the output to the standard output and redirect it from the workflow as before the refactoring?

workflow/rules/map.smk
deliaBlue: If going back to the previous way of returning the output, this argument will be removed. Otherwise, using the new way of specifying the directories on the workflow, the value will be changed to the output directory directly.

.github/workflows/tests.yml
uniqueg: Ah, right, I forgot that the `scripts/` directory _also_ has a `tests/` directory.

But yeah, probably good to have at least some linting on the tests as well, so yeah, maybe open another issue.

scripts/oligomap_output_to_sam_nh_filtered.py
uniqueg: Okay, if you are confident the new script does what the old script does because you tested it, then that's good enough for me in this case. Of course it's better to have extensive unit tests first, before refactoring. But it's not always working like that :)

scripts/oligomap_output_to_sam_nh_filtered.py
uniqueg: Weird. I'm pretty sure though that it's completely different than FASTA (which is an identifier line starting with `>`, followed by one or more lines representing a DNA, RNA or peptide sequence. But for oligomap we get a complex multiline format representing an alignment.

Please resolve the comment after you have changed it.

scripts/oligomap_output_to_sam_nh_filtered.py
uniqueg: Perfect. Please resolve the comment after you are done.

workflow/rules/map.smk
uniqueg: Yes, redirecting the output to `STDOUT` sounds reasonable to me. It's usually the best choice for scripts, because you can use the script in pipes. And it's less code. And overall a more user friendly API, because it requires less specific knowledge on its usage by the user - it will just behave like most other scripts.

Also, you can then name it however you want in the workflow, so no need for the `--ref` as well, I suppose. Besides, if you really want to include a param to take a string to be included in an output file name, consider naming it `--prefix` or something like that. It is much more obvious. Also, set a default.

But anyway, this really only makes sense if you have multiple output files, which should be avoided, if at all possible, for the reasons mentioned above.

workflow/rules/map.smk
uniqueg: Yes, I think this will become much cleaner when you go back to writing to `STDOUT`.

Btw, changing an API is _not_ considered refactoring. Refactoring is when you change the internals of code:
- without changing the interface/API AND
- without changes to any outputs

You can make the code easier to read, create or remove classes and functions, improve the efficiency of the code etc., but from user's perspective, the usage and the outputs should be exactly the same.

scripts/oligomap_output_to_sam_nh_filtered.py
deliaBlue: Yeah, the extension of the output also seemed weird for me but as it seemed to be the way it was set I said nothing.

I will change the output extension on the `oligomap` repository's test data. 
As for the workflow, I'll open an issue to fix that but keep it like that in this PR while changing the docstring data on this script.

scripts/oligomap_output_to_sam_nh_filtered.py
```diff
@@ -307,9 +296,9 @@ def get_sam_fields(aln: list[str]) -> Fields:
     return fields
 
 
-def eval_aln(outfile: Path, nhfilter: int, d: Dict[str, list],
-             minErr_nh: Dict[str, list], fields: Fields) -> None:
-    """Evaluate an alignment to add, discard or write it in the output file.
+def eval_aln(nhfilter: int, d: Dict[str, list], minErr_nh: Dict[str, list],
```
uniqueg, line 299: Python conventions for variable names are to use `snake_case` consistently, not `CamelCase` or a mix of these. So better `min_err_nh` instead of `minErr_nh`.

scripts/oligomap_output_to_sam_nh_filtered.py
```diff
@@ -399,18 +385,11 @@ def eval_aln(outfile: Path, nhfilter: int, d: Dict[str, list],
...
 def main(arguments) -> None:
-    """Convert oligomap alignments FASTA file to SAM file."""
-    outdir = Path(arguments.outdir)
-    outdir.mkdir(parents=True, exist_ok=True)
-    outfile = outdir/f"oligomap_{arguments.ref}_mappings.sam"
-
-    with open(outfile, 'a', encoding="utf-8") as out_file:
-        out_file.write('')
-
+    """Convert the alignments in the oligomap output file to its SAM format."""
```
uniqueg, line 388: Without "its"

uniqueg (approved):
I think this looks good know (apart from two minor things that don't need re-review - so just do the changes, resolve the conversations and then you can merge)

PR #116, opened 2023-10-31
refactor: use `pathlib.Path` instead of `os.path.join`

### LINKED ISSUE #97: Use `pathlib.Path` instead of `os.path.join`
Using `pathlib.Path` instead of `os.path.join` (which implies manipulating strings for path operations) is less verbose, safes and easier to read.

For instance, compare:

```
from pathlib import Path

WF_DIR = Path(workflow.basedir)

...
        WF_DIR / "envs" / "fastx_toolkit.yaml"
```

with

```
import os

        os.path.join(workflow.basedir, "envs", "fastx_toolkit.yaml)
```

The `WF_DIR` instantiation only needs to be done once (and maybe another one for `LOG_DIR` among few others), and each one of the many many lines that do path operations will become much easier to manage.

deliaBlue:
This PR closes #97 .

The usage of `pathlib.Path` instead of `os.path.join` is applied to the main `Snakefile` and the subworkflows `prepare.smk`, `map.smk` and `quantify.smk`.


deliaBlue:
The static code analysis test fails as the PR #114 is not merged yet. Once it is, and merge into this PR, it will pass.

uniqueg (approved):
Nice! So much cleaner - very satisfying :)

PR #118, opened 2023-11-05
test: standardized usage

### LINKED ISSUE #89: Standardize usage, linting and formatting on Snakemake Workfow Catalog
## Problem description

_MIRFLOWZ_ is listed on the [Snakemake Workflow Catalog](https://snakemake.github.io/snakemake-workflow-catalog/).

However, it does not conform to the [standardized usage](https://snakemake.github.io/snakemake-workflow-catalog?rules=true). Also, both [linting](https://snakemake.github.io/snakemake-workflow-catalog/?rules=true#) and [formatting](https://snakemake.github.io/snakemake-workflow-catalog/?rules=true#) currently fail:

image

## To do

Fix:
- [x] Standardized usage
- [x] Linting
- [x] Formatting

## Additional context

See [ZARP](https://snakemake.github.io/snakemake-workflow-catalog?usage=zavolanlab/zarp) for an example:

image

deliaBlue:
This PR closes #89 

The changes in this PR are:

- Add `README.md`to the `config/` directory
- Add format and lint test scripts
- Modify the CI tests to use the scripts
- Add the `.snakemake-workflow-catalog.yml` file
- update path to the `scripts/` directory in `config_schema.json`
- Correct the formatting in `map.smk`
-  Update the main `README.md` to match the explanations of the `config/README.md` along with fixing a couple of typos.


config/README.md
```diff
@@ -0,0 +1,119 @@
...
+together and reproduce an analysis. Assuming that your current directory is
+the repository's root directory, create a directory to store all your data
+and traverse to it with:
+
+```bash
+mkdir path/to/your_run
+cd path/to/your_run
+```
+
+### 1. Prepare the sample table
```
uniqueg, line 48: Is this jumping from a header 1 (`#`) to a header 3 (`###`) required by Snakemake? If not, use header 2 (`##`)

config/README.md
```diff
@@ -0,0 +1,119 @@
+# Dependencies installation
+
+For a faster creation of the environment (and Conda environments in general),
+you can also install [Mamba][mamba] on top of Conda. In that case, replace
+`conda` with `mamba` in the commands below (particularly in 
+`conda env create`).
```
uniqueg, line 6: It's a bit of a weird way to start instructions. I would rather but this as a comment below (and then refer to the "commands above").

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Create and activate the virtual environment with the required dependencies
+with Conda:
+
+```bash
+conda env create -f environment.yml
+conda activate mirflowz
+```
+
+If you plan to run _MIRFLOWZ_ via Conda, we recommend using the following
```
uniqueg, line 16: `[...] creation, especially [...]` rather than `[...] creation specially [...]`

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+In addition, if you want to run _MIRFLOWZ_ via Singularity and do not already
+have it installed globally on your system, you must further update the Conda
+environment with:
+
+```bash
+conda env update -f environment.root.yml
+```
+
+> Mind that you must have the environment activated and root permissions on
+> your system to install Singularity.
```
uniqueg, line 32: You could consider adding this as a subsection `Running _MIRFLOWZ_ with Singularity`.

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+
+# Run the workflow on your own samples
+
+In order to run _MIRFLOWZ_ on your own samples, we recommend having all the
+input files inside a dedicated directory. This way is easier to keep the data
```
uniqueg, line 38: `This way, it is easier [...]`

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+```bash
+touch samples.tsv
+```
+
+> Fill the sample table according to the following requirements:  
+>
+> - `sample`. This column contains the library name.  
+> - `sample_file`. In this column, you must provide the path to the library file.
+> The path must be relative to the working directory.
```
uniqueg, line 62: "working directory" is ambiguous. Is it the Snakemake working directory or the user's current directory (that is, the input file directory in this example)? And if it's the former, how does the user know what's Snakemake's working directory? In my opinion it would be clearer, if the file paths would be relative to the directory that the sample table file is in (but of course that may need a rewrite).

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+> Fill the sample table according to the following requirements:  
+>
+> - `sample`. This column contains the library name.  
+> - `sample_file`. In this column, you must provide the path to the library file.
+> The path must be relative to the working directory.  
+> - `adapter`.  This field must contain the adapter sequence in capital letters.  
+> - `format`. In this field you must state the library format. It can either be 
+> `fa` if providing a FASTA file or `fastq` if the library is a FASTQ file.
```
uniqueg, line 65: I think you don't need to write "This column", "In this field" etc. for each of these, I guess that's really obvious.

My suggestion:
- `sample`: Arbitrary name for the miRNA sequencing library
- `sample_file`: Path to the miRNA sequencing library file. <CONTINUE WITH EXPLANATION OF RELATIVE PATHS; see other comment>
- `adatper`: Sequence of the 3'-end adapter used during library preparation, in capital letters
- `format`: One of `fa` or `fastq`, if the library file is in FASTA or FASTQ format, respetively

> BTW: It's a bit weird/inconsistent that we require `fa` or `fastq` rather than `fa` and `fq`, or `fasta` and `fastq`. Probably it's even better to allow both short and long forms, and independent of case. Always nice to make it easier for the user. And while we are at it, I don't see why we can't just call an `.upper()` on the `adapter` as well, so as to allow users to provide all lower case, all upper case or a mix thereof. Bit annoying to change if they copy the adapter seq from somewhere and it's not in the case we require. Fine with me if you include these changes in _this_ PR, as it is directly related to providing easy instructions here.

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+1. A **`gzip`ped FASTA** file containing **reference sequences**, typically the
+   genome of the source/organism from which the library was extracted.
+
+2. A **`gzip`ped GTF** file with matching **gene annotations** for the
+   reference sequences above.
+
+> _MIRFLOWZ_ expects both the reference sequence and gene annotation files to
+> follow [Ensembl][ensembl] style/formatting. If you obtained these files from
+> a source other than Ensembl, you may first need to convert them to the
+> expected style to avoid issues!
```
uniqueg, line 80: Would be cool if you found a reference to point people to, like a tool or a tutorial or something like that. But don't break your neck over it.

uniqueg: Also, I would formulate it stronger:

"If you obtained these files from a source other than Ensembl, you must ensure that they adhere to the expected format by converting them, if necessary."

Same below.

config/README.md
```diff
@@ -0,0 +1,119 @@
...
+[configuration file template](config_template.yaml).
+
+```bash
+cp ../config/config_template.yaml config.yaml
+
+```
+
+Open the new copy in your editor of choice and adjust the configuration
+parameters to your liking. The template explains what each of the parameters
+means and how you can meaningfully adjust them.
```
uniqueg, line 113: "mean" instead of "means"

config/README.md
uniqueg, line 1: Nicely written :)

uniqueg (changes requested):
A few comments. And of course also the integration tests currently fail. Any idea why the three output files change?

config/README.md
deliaBlue: I've taken most of the text from the main README and modify it to make sense in the required context. 

This makes me wonder... should I apply the same changes you are suggesting to the mail README?

config/README.md
deliaBlue: The solution I found to allow the end-user to use both the short and the long version of the FASTA and FASTQ files is to duplicate the rules that depends on it using the different versions of the file extension ( see [stackOverflow answer](https://stackoverflow.com/a/56408220)

uniqueg: No. There must be a better way. It's just a string param. I'm pretty sure Snakemake allows applying functions to parameters, so you should be able to write code that converts it to whatever you need, while still allowing users to provide something more simple/intuitive/forgiving. CJ should know about this, she has done something similar for ZARP.

uniqueg (approved):
I guess this is fine now? So really the change of Singularity and Conda resolved the failing tests? So weird...

PR #124, opened 2023-11-16
test: fix workflow descriptors format and lint test

deliaBlue:
This PR has three major changes:

- remove the CLI flag `-l 80` from the format test script
- modifiy the Snakemake files to pass the format check
- add the singularity flags in the lint test
- create `config_lint.yaml` for the linting test
- add path to the new config file in the Snakefile

### DIFF
--- a/test/config_lint.yaml
+++ b/test/config_lint.yaml
@@ -1,0 +1,6 @@
+samples: test/test_files/samples_table.tsv  
+genome_file: test/test_files/genome.fa.gz
+gtf_file: test/test_files/gene_annotations.gtf.gz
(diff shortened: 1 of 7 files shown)


uniqueg (dismissed):
I'm approving this because I'm totally cool with defining a default config file in the workflow like you did. It could be useful even outside the Snakemake Workflow Catalog use case just for people to play with the workflow without/before preparing actual inputs in a "real" config file.

Whether you wanna look into defining a mandatory flag for the config file, I leave it up to you. There are arguments for and against it, but I don't think it's all that important, so I'm fine with either.

PR #125, opened 2023-11-21
style:  change unit test scripts for `flake8`

### LINKED ISSUE #120: refactor: static-code analysis for the unit test scripts
Modify the unit-test scripts to pass the static code analysis.
In addition, modify the CI rules in order to check both the scripts and the unit tests.

deliaBlue:
This PR closes #120 .

All the unit test scripts are now passing the `flake8` analysis.

Moreover, the call for ''flake8'' in the CI has been changed. 
For each unit test script,  the error ` E402 module level import not at top of file` appears. Given that the modules are found in the parent directory, to set the path there is required and done after importing `sys`.  Hence, the addition of `--ignore=E402`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: flake8
         working-directory: ./scripts
-        run: flake8 ./*.py
+        run: flake8 --ignore=E402
 
       - name: mypy
         working-directory: ./scripts
--- a/scripts/tests/test_filter_multimappers.py
+++ b/scripts/tests/test_filter_multimappers.py
@@ -22,7 +22,7 @@ from scripts.filter_multimappers import (
 def sam_empty_file():
     """Import path to empty test file."""
     empty_file = Path("files/header_only.sam")
-    
+
     return empty_file
 
 
@@ -42,6 +42,7 @@ def sam_no_multimappers_file():
 
     return no_multi
 
+
 @pytest.fixture
 def sam_unique_diff_multimappers_files():
     """Import path to test files with a single multimapper."""
@@ -49,6 +50,7 @@ def sam_unique_diff_multimappers_files():
     out_diff_multi = Path("files/diff_multimappers.sam")
 
     return in_diff_multi, out_diff_multi
+
 
 @pytest.fixture
 def sam_unique_equal_multimapper_files():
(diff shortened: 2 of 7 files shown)


deliaBlue:
By ignoring the error `E402`, all python scripts failed the static code analysis. 
After fixing them, an extra error is add to be ignore, error `W504`. 

This error does not allow to break a line after/before an arithmetic operator enabling the possibility of concatenating strings in different lines (which avoids having >79 char long lines).

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: flake8
         working-directory: ./scripts
-        run: flake8 ./*.py
+        run: flake8 --ignore E402,W504
```
uniqueg, line 41: Could you put these flags/rules into a config file maybe? For example in a `setup.cfg` file in the root directory under header `[flake8]`?

uniqueg: On the other hand though, do we really need to have an import that somewhere other than the top of the file? Smells like bad code design. In any case, we should most certainly not ignore [`E402`](https://www.flake8rules.com/rules/E402.html) warnings across the entire project. If we really can't (easily) avoid it, just ignore it inline in the place where that warning is raised (I hope it's only one).

Similarly, I'm pretty sure [`W504`](https://www.flake8rules.com/rules/W504.html) can always be avoided?

scripts/filter_multimappers.py
```diff
@@ -128,17 +128,16 @@ def find_best_alignments(
 
     aln_indels = [(aln, count_indels(aln=aln)) for aln in alns]
     min_indels = min(aln_indels, key=lambda x: x[1])[1]
-    best_alignments = [
-            aln
-            for i, (aln, indels) in enumerate(aln_indels)
-            if indels == min_indels]
+    best_alignments = [aln
+                       for i, (aln, indels) in enumerate(aln_indels)
+                       if indels == min_indels]
```
uniqueg, line 133: Was this done by `black`?

scripts/mirna_quantification.py
```diff
@@ -285,16 +285,18 @@ def collapsed_contribution(aln: pysam.AlignedSegment) -> float:
...
-                         " not include the number of collapsed sequences.\n" +
-                         "Check SAM file consistency and CLI options" +
-                         " --collapsed and --nh.\n")
+        sys.stdout.write(
+            f"Invalid query name: '{aln.query_name}'.\n" +
+            "Option --collapsed specified but query name does " +
+            "not include the number of collapsed sequences.\n" +
+            "Check SAM file consistency and CLI options" +
+            " --collapsed and --nh.\n"
+        )
```
uniqueg, line 294: What's with the all the `+`es? Better to use parantheses for multiline strings. Or triple quoted strings. Weird that none of the linters complain here.

uniqueg (changes requested):

scripts/filter_multimappers.py
uniqueg: Okay. But I would actually recommend to use 'black' and include a `black --check` call in the CI. That way, we will always have the same general style without any hassles.

uniqueg (approved):
Nice :)

Issue #126, opened 2023-11-30
docs: describe workflow rationale


deliaBlue:
The user should be able to decide if _MIRFLOWZ_ is the correct tool for them to use by reading how it works in the documentation. 

To achieve so extend the explanation in the main `README.md`  to include details on:

- [ ] how the counting is done
- [ ] how the mapping and filtering are done
- [ ] the output files the user will get with some examples

In addition, in the `pipeline_documentation.md`, include examples of the common and edge cases that highlight the workflow behavior.

PR #129, opened 2023-12-02
feat: set intermediate and final output files

### LINKED ISSUE #86: Clean/order final output files in `Snakefile`
Once the pipeline is done, check which final output files are really needed/wanted and make the required changes.

deliaBlue:
Fixes #86

- add  a directory to store the intermediate files in `config_template.yaml` and `config_schema.json`
- add the intermediate files path in all the snakemake files
- add the `onfailure` and `onsuccess`  handlers in the main Snakefile so intermediate files are not lost if the workflow fails and removed otherwise.
- redirect the intermediate files to the new directory in the sub-workflows
- add the section _"Expected output files"_ in the main `README.md` with a brief explanation of the files that are kept by default and how to keep the intermediate ones
- update the `finish` rules in `pipeline_documentaion.md`
- update  the rule graph
- add the CLI option `--no-hooks` to all the tests in order to keep checking all the files in the CI tests
- update the expected output to match the new location of the intermediate files


README.md
```diff
@@ -251,6 +252,50 @@ snakemake \
...
+the final counting.
+
+4. A BAM file containing the uncollapsed set of alignments contributing to the
+final counting and its corresponding index file (`bam..bai`).
+
+5. Table(s) containing the counting data from all libraries for (iso)miRs
+and/or pri-miRs. Each row corresponds to a miRNA species, and each column
+represents a sample library. Counting involves aggregating contributions from
+all alignments, calculated as the ratio of collapsed reads in th alignment to
+the number of hits (NH value).
```
uniqueg, line 280: I think we take care of collapsing, so I don't think you need to mention it here. Rather, speak from the perspective of a single read. Also, the sentence is hard to understand and contains a typo. Better to say something like:

> Each read is counted towards all the miRNA species it aligns to, with 1 / n, where n is the number of genomic or transcrptomic loci that the read aligns to.

I'm not actually this is correct (is it genomic _and_ transcriptomic? do we really consider _all_ such loci, or only those for which miRs are annotated?), but I'm sure you will know it better and can adapt accordingly.

README.md
```diff
@@ -251,6 +252,50 @@ snakemake \
...
+    --configfile="path/to/config.yaml" \
+    --use-conda \
+    --printshellcmds \
+    --rerun-incomplete \
+    --no-hooks \
+    --verbose
+```
+
+After successful execution of the workflow, the intermediate files will be
+found in the `results/inter_files` directory.
```
uniqueg, line 297: I generally think it's good to name things explicitly. So perhaps rather name the directory `intermediate_files` or `intermediates`, rather than `inter_files`.

README.md
```diff
@@ -251,6 +252,50 @@ snakemake \
 After successful execution of the workflow, results and logs will be found in
 the `results/` and `logs/` directories, respectively.
 
+### Expected output files
```
uniqueg, line 255: Nice!

config/config_schema.json
```diff
@@ -30,6 +30,11 @@
             "default": "results/",
             "description": "Path to the output directory."
         },
+        "tmp_dir":{
+            "type": "string",
+            "default": "results/inter_files",
+            "description": "Path to the temporary directory storing the intermediate files."
+        },
```
uniqueg, line 37: I think this may potentially be confusing. Temporary files, in the common usage of the term, are files generated by a single program. Indeed, some of the tools that we will use will generated temporary files, and these will not necessarily be stored in the same location (but rather whatever the `TMPDIR` and `TMP` environment variables point to).

config/config_schema.json
```diff
@@ -30,6 +30,11 @@
             "default": "results/",
             "description": "Path to the output directory."
         },
+        "tmp_dir":{
+            "type": "string",
+            "default": "results/inter_files",
+            "description": "Path to the temporary directory storing the intermediate files."
```
uniqueg, line 36: > [...] to the ~temporary~ directory [...]

workflow/Snakefile
```diff
@@ -31,6 +31,23 @@ validate(config, Path("../config/config_schema.json"))
 
 
 OUT_DIR = Path(config["output_dir"])
+TMP_DIR = Path(config["tmp_dir"])
+LOG_DIR = Path(f"{config['local_log']}/../")
+
+
+###############################################################################
+### onSuccess/onError handlers configuration
+###############################################################################
```
uniqueg, line 40: Nice!

workflow/Snakefile
```diff
@@ -31,6 +31,23 @@ validate(config, Path("../config/config_schema.json"))
...
+
+
+onsuccess:
+    print("\nWORKFLOW SUCCEED. Removing intermediate files.\n")
+    shell("rm -rf {TMP_DIR}")
+
+
+onerror:
+    print("\nWORKFLOW FAILED. Check the log file in the LOGS/ directory.\n")
```
uniqueg, line 49: Is this the actual name of the directory, `LOGS/`? If not, just say "log directory", because the capitalization and trailing slash give the impression that it is indeed the name of the directory.

uniqueg (dismissed):
Good work :) Couple of comments, but shouldn't be all that much

workflow/Snakefile
deliaBlue: It is in fact the `LOGS/` directory. I write it that way because it its in that directory whereas all the rule logs are in the `LOGS/LOCAL/` or `LOGS/CLUSTER` directory. Still, I can change it if you think that's better.

README.md
```diff
@@ -251,6 +252,50 @@ snakemake \
...
+
+Upon successful execution of _MIRFLOWZ_, the tool automatically removes all
+intermediate files generated during the process. The final output comprises:
+
+1. A SAM file containing alignments intersecting a pri-miR locus. These
+alignments intersect with extended start and/or end positions specified in the
+provided pri-miR annotations. Please note that they may not contribute to the
+final counting and will not appear in the final table.
```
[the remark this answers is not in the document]
deliaBlue: Should I add this line here? I mean for this PR sure, but issue #126 could in fact, explain it in more detail. Then I could return this explanation into its current format and redirect the user to the section/file if it wants more information.

workflow/Snakefile
uniqueg: If it always `LOGS/`, i.e., users can't change it through a config param, you can leave it. Otherwise it may be better to be more generic

uniqueg (dismissed):
I'll approve this now. You can see how/whether you address (or not) the two open conversations and then merge.

PR #130, opened 2023-12-08
fix: suffix nomenclature for SAM uncollapsing 

### LINKED ISSUE #127: fix: empty SAM file after reads uncollapsing
Once the counting is done and the final alignments are to be uncollapsed, the resulting SAM file contains only the header without alignments regardless of reads being or not collapsed. 

The script uncollapsing the reads in the SAM file is [sam_uncollapse.pl](https://github.com/zavolanlab/mirflowz/blob/dev/scripts/sam_uncollapse.pl).

deliaBlue:
Fixes #127 .

This PR refactors [sam_uncollapse.pl](scripts/sam_uncollapse.pl) to correctly uncollapse alignments regardless of the suffix format assuming the main format to be the one done by FASTX TOOLKIT. 
To achieve this, the regex call now captures and stores the first element prior the dash (`-`) as the alignment's name to be used and the integer that follows as the number of collapsed reads. This change keeps the previous functionality of the script while allowing the QNAME to have extra characters after the amount of collapsed reads and correctly uncollapse the SAM file.

In addition, it updates the expected output with the correct uncollapsed SAM, BAM and BAI files.


scripts/sam_uncollapse.pl
```diff
@@ -128,15 +128,16 @@ sub sam_uncollapse {
...
-                        my $repeat = $1;
+                        my $read = $1;
+                        my $repeat = $2;
 			# If --suffix option is set...
 			if ($suffix) {
 				# Iterate over number of identical reads/alignments
 				for my $suffix (1..$repeat) {
 		                        # Recreate line with suffix
-                		        $line = join "\t", "$id.$suffix", $rest;
+                		        $line = join "\t", "$read.$suffix", $rest;
```
uniqueg, line 140: At a quick glance, this actually looks quite different from the previous behavior. So different, in fact, that it might be a breaking change, i.e., that old calls of the script fed old inputs (with the expectations we had on the inputs back then) would result in different outputs. If that is indeed the case, I think it would be better to find a solution that is _not_ breaking, i.e., the script should behave in exactly the same way if faced with old inputs, but _also_ should be able to process the new input ID format according to our needs. If necessary, you could add a command-line option. But perhaps you can also manage without (and instead determine the input format and associated expected behavior automatically).

It would also be good if the PR description contained some concrete examples that highlight how the previous behavior (on old inputs) is unchanged, and how the change affects only new inputs.

scripts/sam_uncollapse.pl
```diff
@@ -92,7 +92,7 @@ sub usage {
 Comments:
 CAUTION: Only marginal validation of the input file type/format performed!
 
-Version 1.2 (2014-08-26)
+Version 1.2.1 (2023-12-08)
```
uniqueg, line 95: This would at least constitute a minor version change, if not a major version change (see next comment on breaking changes).

uniqueg (changes requested):

uniqueg (approved):

PR #137, opened 2024-02-06
build: set up Black correctly and Snakemake format

deliaBlue:
This PR:

- corrects the `Black` test in the CI file
- adds `Black` in  `environment.dev.yml` with its current version as the minimum one
- formats the main Snakefile with `snakefmt`
- adds `git-blame-ignore-revs` with the `snakefmt` commit

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,12 +49,8 @@ jobs:
         run: pylint --rcfile=../pylint.cfg ./*.py
 
       - name: black
-        uses: psf/black@stable
-        with:
-          options: "--check --verbose --line-length=79"
-          src: "./scripts"
-          version: "24.1.1"
-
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -51,8 +51,9 @@ jobs:
       - name: black
         uses: psf/black@stable
         with:
-          options: "--check --verbose --line-length=79 --required-version 24.1.1"
+          options: "--check --verbose --line-length=79"
           src: "./scripts"
+          version: "24.1.1"
```
uniqueg, line 56: But doesn't that mean that everytime we upgrade Black, we have to change the version in two places? 🤔

uniqueg (approved):
Approving already as I'm fine with merging either way, but would prefer to have only one source of Black version, if possible

PR #139, opened 2024-02-07
fix: add missing "conda" directive

### LINKED ISSUE #138: fix: cannot compress with `bgzip` using Conda
The rule `compress_reference_genome` in `pileup.smk` assumes the end user can compress the genome with `bgzip`  when running the workflow with Conda.

Add a Conda environment on that rule to make sure the rule can be executed and the pipeline is not terminated.

deliaBlue:
This PR closes #138 .

In order to use `bgzip` when running the pipeline with Conda, the SAMtools micro-environment is used in the `compress_reference_genome` rule in `pileup.smk`


deliaBlue:
The format test does not pass as the fix is done in the PR #137 and it is not yet merged.

uniqueg (changes requested):
I'm not sure how the proposed changes match the title of this PR or the referenced issue (which, admittedly, I don't quite understand)

deliaBlue:
The issue behind this PR was opened because when trying to use `bgzip` when running the pipeline using Conda, the command could not be found - I have it installed locally and I thought it was on every machine.

workflow/rules/quantify.smk
```diff
@@ -64,7 +63,10 @@ rule finish_quantify:
     input:
         primir_intersect_sam=OUT_DIR / "{sample}" / "alignments_intersecting_primir.sam",
         mirna_intersect_sam=OUT_DIR / "{sample}" / "alignments_intersecting_mirna.sam",
-        table=OUT_DIR / "TABLES" / "all_{mir}_counts.tab",
+        table=expand(
+            OUT_DIR / "TABLES" / "all_{mir}_counts.tab",
+            mir=[mir for mir in config["mir_list"] if mir != "isomir"],
```
uniqueg, line 68: But this looks like something unrelated to the environment issue? Are you trying to sneak something past me here? :raised_eyebrow::laughing: 

Or is this some artifact from another PR?

uniqueg (approved):

PR #140, opened 2024-02-11
fix: allow empty input tables

### LINKED ISSUE #136: fix: `merge_tables.R` fails if given empty input
When any of the input files provided to `merge_tables.R`, the rule fails and the workflow stops. The error message is:

```console
Error in read.table(dataFiles[1], sep = "\t") : 
  no lines available in input
Calls: merge_tables -> read.table
Execution halted
```

Fix the script so it can take an empty input file and add it as column full of `NA`s in the final table.

deliaBlue:
This PR closes #136 .

The changes in `merge_tables.R` are:

- Add a new function `get_table()` that tries to read the input file and if empty, create a data frame made of one row with `NA`s in each column
- Refactor the `merge_tables()` function to include the new function and, if an empty file was present, remove the row with a `NA` in the miRNA species column
- Document both functions
- Refactor the code (max. line length, space between operators...) according to the R style
- Update the description message


deliaBlue:
The format test does not pass as the fix is done in the PR #137  and it is not yet merged

scripts/merge_tables.R
```diff
@@ -73,7 +78,18 @@ option_list <- list(
...
+    usage = paste(
+      "Usage:",
+      script,
+      "[OPTIONS] --input_dir <path/to/input/files>\n",
+      sep = " "
+    ),
+    option_list = option_list,
+    add_help_option = FALSE,
+    description = msg
+  )
```
uniqueg, line 92: I'm not sure which linter you have used? I'm fine with all the line wraps, but there are some issues. For example, I am pretty sure it's convention that there is no white space between function args, the `=` and the value, i.e., it should be `arg=value`, not `arg = value`.

Some more comments on this later, probably.

scripts/merge_tables.R
```diff
@@ -7,8 +7,12 @@
...
 
 # Import required packages
-if ( suppressWarnings(suppressPackageStartupMessages(require("optparse"))) == FALSE ) { stop("[ERROR] Package 'optparse' required! Aborted.") }
-if ( suppressWarnings(suppressPackageStartupMessages(require("dplyr"))) == FALSE ) { stop("[ERROR] Package 'dplyr' required! Aborted.") }
+if (suppressWarnings(suppressPackageStartupMessages(require("optparse"))) == FALSE) {
+  stop("[ERROR] Package 'optparse' required! Aborted.")
+}
+if (suppressWarnings(suppressPackageStartupMessages(require("dplyr"))) == FALSE) {
+  stop("[ERROR] Package 'dplyr' required! Aborted.")
+}
```
uniqueg, line 15: Refactoring/linting was not really the point of this issue. It's best not to mix these up, as it will make it harder to focus on the changes to fix an issue, as well as for reviewing. But okay in this case, as it's not that much...

scripts/merge_tables.R
```diff
@@ -21,9 +25,10 @@ script <- sub("--file=", "", basename(commandArgs(trailingOnly=FALSE)[4]))
 # Build description message
 description <- "Merge miRNAs quantification tables.\n"
 author <- "Author: Paula Iborra, Biozentrum, University of Basel"
-version <- "Version: 1.0.0 (JUN-2019)"
-requirements <- "Requires: optparse"
-msg <- paste(description, author, version, requirements, sep="\n")
+mantainer <- "Refactor and documentation: Iris Mestres"
+version <- "Version: 1.1.0 (FEB-2024)"
+requirements <- "Requires: optparse, dplyr"
```
uniqueg, line 30: Good catch!

scripts/merge_tables.R
```diff
@@ -21,9 +25,10 @@ script <- sub("--file=", "", basename(commandArgs(trailingOnly=FALSE)[4]))
 # Build description message
 description <- "Merge miRNAs quantification tables.\n"
 author <- "Author: Paula Iborra, Biozentrum, University of Basel"
-version <- "Version: 1.0.0 (JUN-2019)"
-requirements <- "Requires: optparse"
-msg <- paste(description, author, version, requirements, sep="\n")
+mantainer <- "Refactor and documentation: Iris Mestres"
```
uniqueg, line 28: You could put this as just:

`maintainer: Iris Mestres <your@email.address>`

without the `Refactor and documentation`.

And while you are at it, perhaps change this one as well:

`author: Zavolan Lab <zavolab-biozentrum@unibas.ch>`

If you prefer, you could also set both to `zavolab-biozentrum@unibas.ch`.

scripts/merge_tables.R
```diff
@@ -113,22 +190,35 @@ merge_tables <- function(cwd, prefix){
...
 # Write log
-if ( verb ) cat("Creating output directory...\n", sep="")
+if (verb)
+  cat("Creating output directory...\n", sep = "")
 
 # Create output directories
-dir.create(dirname(out.file), recursive=TRUE, showWarnings=FALSE)
+dir.create(dirname(out.file),
+           recursive = TRUE,
+           showWarnings = FALSE)
```
uniqueg, line 199: Not a big fan of this line wrapping style. But more importantly, it's inconsistent with the one you use later, on lines 213-220 (which I prefer).

You can pick whichever one you prefer (regardless of my preference), but make sure it's being used consistently.

scripts/merge_tables.R
```diff
@@ -92,19 +108,80 @@ if ( is.null(in.dir) ) {
...
+      colnames(table) <- fields
+      return(table)
+    }
+  )
+}
+
+#' Merge tables with the same prefix
+#'
+#' `merge_tables()` takes all the files in `cwd` that start with `prefix` and
+#' merge them keeping all the miRNA species present in each of the tables.
```
uniqueg, line 150: `[...] merges them, keeping [...]` (plural)

scripts/merge_tables.R
```diff
@@ -92,19 +108,80 @@ if ( is.null(in.dir) ) {
...
+#'
+#' @details The function `get_table()` is used to make sure that even if an
+#'   empty input file is given, the merge can still be done. Thus, before
+#'   returning the merged table, the row with a `NA` in the `ID` field, if any,
+#'   is removed.
+#'
+#'   The function `dplyr::full_join()` method is used for the merge, therefore,
+#'   if a miRNA species in `ID` is missing in any of the tables being joined,
+#'   its value is set to `NA` in that column.
```
uniqueg, line 159: I think this description is a bit confused/confusing. It would probably make sense to first describe why and under what conditions rows with `NA` in the `ID` field are added before you say that they are removed. Also, both the `Thus, ` and `therefore, ` seem to not describe something that is logically following (at least not immediately/obviously), so I guess some other words should be used here.

scripts/merge_tables.R
```diff
@@ -92,19 +108,80 @@ if ( is.null(in.dir) ) {
...
+#' @param prefix String to be removed from the input file name. It must be
+#'   present in all the tables to be merged.
+#'
+#' @returns `get_table()` returns a data frame containing the miRNA species to
+#'   be counted in first column, named `ID`, and their counts in that file in
+#'   the second one. The name of the second column in the data frame is obtained
+#'   by removing the `prefix` from the input file name. If no `prefix` is given,
+#'   the whole file name is used.
+#'
+#'   If the input file is empty, the returned data frame will consist on one row
```
uniqueg, line 126: `consist of`

uniqueg (changes requested):
Looks good, just some minor issues :)

scripts/merge_tables.R
deliaBlue: I used the code format  tool within Rstudio. Now I used [lintr](https://lintr.r-lib.org/) with the default style `"tidyverse"`, and in fact, it made me put spaces between the `=` and whatever there was before and after...

uniqueg: Fine with me. Doesn't matter if I like it as long as things are consistent :) But wondering why it was wrapping lines in different styles. Anyway, not worth investigating.

uniqueg (approved):
It's actually really nice with the linting and docstrings and all of that good stuff. Good job! :+1:

PR #142, opened 2024-05-02
fix(mirna_extension): set correct coordinates

deliaBlue:
This PR fixes the static code analysis test by refactoring the script `mirna_extension.py`.
The changes are:

- Use `min()` and `max()` to asses the final start and end positions of the precursor miRNA.
- The casting to `int` has been removed to avoid possible value changes.
- The variable `mir` has been changed to `mature_mir` for clarity.

Finally, the expected output has been updated due to the extension values to differ. 
In the previous version of the script, original pri-miR coordinates were used to decide if the annotated coordinates were to be modified _in situ_ but that variable was not updated. This lead to always compare with a higher value (for example 121035) than the previously modified one (for example 121032) and therefore, ending up with an incorrect value (for example, 121034 < 121035 but 121034 > 121032).

In this version of the script, the variable used to compare is updated (if necessary) in each iteration and the annotated pri-miR coordinates are updated at the end of the loop with the lowest/higher value.

```console
PREVIOUS START POSITION MODIFICATIONS:
121035 -> 121032 -> 121034

CURRENT START POSITION MODIFICATIONS:
121035 -> 121032 -> 121032
```


test/expected_output.md5
```diff
@@ -54,7 +54,7 @@ f5749f5d2096386444a3d021e0a01817  results/intermediates/genome_processed.fa.bz.g
 d34fc868b861b1bc46db07a397dc0f10  results/intermediates/genome_processed.fa.fai
 21e102e4ebd3508bb06f46366a3d578d  results/intermediates/exons.gtf
 003b92b245ac336e3d70a513033e1cee  results/intermediates/transcriptome_trimmed_id.fa
-cc5c3512dab0e269d82bd625de74198e  results/intermediates/extended_primir_annotation_6_nt.gff3
```
uniqueg, line 57: It's not obvious to me why the output changes? Can you please elaborate on that?

uniqueg: I have just seen that you wrote in the PR description that this is
> due to the extension values to differ

If things like that happen, please describe in the PR description exactly how results change and why. If the implementation fixes a bug, this is certainly not a `ci:` issue anymore, but `fix:`.

Also, I think it should anyway be a `refactor:` rather than a CI, because you are refactoring (if not fixing) the code because of a CI issue, but you aren't actually updating the CI at all.

deliaBlue: As we previously talked, this script requires another line of code to correctly extend the miRNA primary transcripts in relation to the mature miRNA extended coordinates (one that accounts for the strandness). The previous expected output is not correct and neither is the updated one. Why they differ is still ambiguous to me even if the reason why it is wrong is the same.

You suggested to firstly fix the static code analysis error in a PR and make another one with the extra line of code. Now, if this can PR can be a refactor regarding the correct way to extend pri-miRs, I could add this extra line and then provide a clear explanation of why the expected output is different.

uniqueg: IIRC correctly, the issue you are trying to address here is that the CI just started failing, but not in response to a code change, but rather because of some different environment being set up because of a different version of a dependency or sth like that. To do this, the idea would be to fix it equivalently, i.e., `refactor` the code, not `fix` a bug. Either way it's not a `ci` commit, because the CI pipeline doesn't change, only its output (hopefully). What should _not_ change is the output though, unless it is, indeed, a fix.

Now, normally, if you have to fix basically the same code _anyway_, I would say, okay, let's just go ahead and fix it in a way that makes the CI issue go away. But in this case, I feel that we want to know what is going on, because otherwise maybe your change will only mask another underlying (and unidentified) issue and we will never know, because once you fix, we do _not_ expect the same outcome, and if it changes we may not look at the differences closely and therefore not realize that the output may have actually changed not because of the fix but something else. Or both. 

I think that if there are differences in the outputs, excepted or unexpected, it would make sense to explore what changed, and whether it changed in the way we expect (if we do expect it). For now you can do this manually with a `diff`, but it's probably worth it to automate that in a future (`ci`) PR, where you could add a little script or one-liner that calculates the MD5 sums as usual, but which automatically shows a `diff` in the log file for every file that changed.

uniqueg: I am not sure why you have closed this conversation. Maybe you didn't see my last comment?

The PR currently now looks good, as in the name of the PR and the proposed changes match, because now you _do_ fix something and therefore we expect a different output file. And from what I can see, the implementation should now be fine.

However, it is fishy that the previous change - which was supposed to be only a refactoring - changed the output. Couldn't you just run a diff on that version to see what was going on? It would give me peace of mind with respect to what I wrote in my last comment above.

deliaBlue: I did run a _diff_ on the current implementation and the one in this PR and found out the reason behind the change in the output.I've delete the files and I'd have to change branches, run tests, moving files...  So I'll try to explain it with the actual affected line but handmade - hope it serves for you.

In the reference annotation file, there are the following relevant entries:

```
chr19	.	miRNA_primary_transcript	121035	121101	.	+	.	ID=MI0000779;Alias=MI0000779;Name=hsa-mir-371a
chr19	.	miRNA	121040	121059	.	+	.	ID=MIMAT0004687;Alias=MIMAT0004687;Name=hsa-miR-371a-5p;Derives_from=MI0000779
chr19	.	miRNA	121076	121098	.	+	.	ID=MIMAT0000723;Alias=MIMAT0000723;Name=hsa-miR-371a-3p;Derives_from=MI0000779
chr19	.	miRNA_primary_transcript	121037	121102	.	-	.	ID=MI0017393;Alias=MI0017393;Name=hsa-mir-371b
chr19	.	miRNA	121075	121096	.	-	.	ID=MIMAT0019892;Alias=MIMAT0019892;Name=hsa-miR-371b-5p;Derives_from=MI0017393
chr19	.	miRNA	121038	121060	.	-	.	ID=MIMAT0019893;Alias=MIMAT0019893;Name=hsa-miR-371b-3p;Derives_from=MI0017393
```

Both mature miRNAs for _hsa-mir-371a_ and _hsa-mir-371b_ are within the range of both the miRNA primary transcript but in different strands. Under the scenario in which the strand is not accounted when picking the mature miRNA of a particular pri-miR, the four annotations are used to extend both species.

As mention in the PR description, the extended annotation is always compared with the original one but the pri-miR coordinates are updated each time. In each iteration, the pri-miR annotations change in the following fashion:

```
ITERATION 1
chr19	.	miRNA_primary_transcript	121032	121101	.	+	.	ID=MI0000779;Alias=MI0000779;Name=hsa-mir-371a
chr19	.	miRNA_primary_transcript	121032	121102	.	-	.	ID=MI0017393;Alias=MI0017393;Name=hsa-mir-371b

ITERATION 2
chr19	.	miRNA_primary_transcript	121034	121101	.	+	.	ID=MI0000779;Alias=MI0000779;Name=hsa-mir-371a
chr19	.	miRNA_primary_transcript	121034	121102	.	-	.	ID=MI0017393;Alias=MI0017393;Name=hsa-mir-371b

ITERATION 3
chr19	.	miRNA_primary_transcript	121034	121102	.	+	.	ID=MI0000779;Alias=MI0000779;Name=hsa-mir-371a
chr19	.	miRNA_primary_transcript	121034	121102	.	-	.	ID=MI0017393;Alias=MI0017393;Name=hsa-mir-371b

ITERATION 4 AND FINAL ANNOTATIONS
chr19	.	miRNA_primary_transcript	121034	121104	.	+	.	ID=MI0000779;Alias=MI0000779;Name=hsa-mir-371a
chr19	.	miRNA_primary_transcript	121034	121104	.	-	.	ID=MI0017393;Alias=MI0017393;Name=hsa-mir-371b
```

In the refactoring, as the comparison is made with the _per-iteration_ updated value and the pri-miR are updated at the end of the loop, the resulting annotations are:

```
chr19	.	miRNA_primary_transcript	121032	121104	.	+	.	ID=MI0000779;Alias=MI0000779;Name=hsa-mir-371a
chr19	.	miRNA_primary_transcript	121032	121104	.	-	.	ID=MI0017393;Alias=MI0017393;Name=hsa-mir-371b
```

It is important to stress the fact that the four mature miRNAs are used as otherwise, both pri-miR would have different extended coordinates.

I hope this makes it clear. I will leave the comment unsolved for you to close.

uniqueg: Hi @deliaBlue. I took the liberty to refactor the (entire) script to make it more readable. See the attached PR. It would still need you to update the tests accordingly, but once that is done and merged into this branch, I think we can merge this PR.

PR #144, opened 2024-05-19
fix: fix & refactor miRNA extension script

uniqueg:
I had a bit of time and felt like coding, so I refactored the script a bit (well, maybe a bit more than a bit).

It should now do what it's supposed to do, I hope. Or in any case, it's broken down into more, simpler steps, so if it isn't fully there yet (the tests would show), then it should be easy enough to change.

I suppose you could merge this in your branch, if you like, and then update the tests to match and cover all edge cases. Note that I have included two or three specific error checks that were not covered before, so apart from refactoring, you'd probably need to add a couple more tests to check for these.


uniqueg:
Of course you can also update/add the tests _first_, in this branch, so that the tests pass. Probably makes more sense :)

scripts/mirna_extension.py
```diff
@@ -1,74 +1,118 @@
...
-        db:
-            in-memory database of the GFF3 file.
-
-    Methods:
-        load_gff_file:
-            creates an in-memory db from an input file
-        extend_mirnas:
-            extends the start and end of the mature miRNAs by n nucleotides
+        db: In-memory database of the input GFF3 records.
+        db: In-memory database of the updated GFF3 records.
```
deliaBlue, line 16: Should one of these be `db_out`?

scripts/mirna_extension.py
```diff
@@ -1,74 +1,118 @@
...
+                        self.seq_lengths[ref_seq] = int(length)
+                    except ValueError as exc:
+                        raise ValueError(
+                            f'Invalid length: "{length}" for sequence "{ref_seq}";'
+                            ' integer value expected'
+                        ) from exc
+
+    def process_precursor(
+        self,
+        precursor: gffutils.Feature,
```
deliaBlue, line 69: There's no doctstring in this method

scripts/mirna_extension.py
```diff
@@ -79,159 +123,132 @@ def extend_mirnas(
...
-    """Command-line arguments parser."""
+    """Parse command-line arguments."""
+    description = """Extend miRNA start and end coordinates.
+
+Extends mature miRNA start and end coordinates in a GFF3 file by the indicated number of
+nucleotides, but not exceeding chromosome boundaries. If the extension causes the mature
+miRNA coordinates to exceed the boundaries of the corresponding precursor (see note in
+"Constraints" below), the precursor coordinates will be extended accordingly. In this
+case, the precursor name will be appended with "_-n" and/or "_+m", where n and m
+represent the extensions on the 5' and 3' end, respectively. The modified mature miRNA
```
deliaBlue, line 174: I don't think the "In this case, the precursor name will be appended with ..." statement is accurate. Regardless of the extension the precursor name be modified. If no extension is made, the name will end in "_-0_+0".

deliaBlue: Also, the whole description has no indentation.

scripts/mirna_extension.py
```diff
@@ -79,159 +123,132 @@ def extend_mirnas(
...
+        type=Path,
+        default=Path.cwd(),
+        metavar="PATH",
     )
-
     return parser
 
 
-def main(arguments) -> None:
+def main(args) -> None:
```
deliaBlue, line 237: I believe the usage of "arguments" instead of "args" was due a static code analysis failing if doing so.

deliaBlue (changes requested):
Here are some small comments; besides the last one, everything is related to the docstrings . 
In the meantime, if it's alright, I'll be working on fixing the unit tests and static code analyses for this PR.

scripts/mirna_extension.py
deliaBlue: Here's the error from pylint:

```
mirna_extension.py:247:9: W0621: Redefining name 'args' from outer scope (line 268) (redefined-outer-name)
```

I can change `args` in the last two lines instead of in the main function

scripts/mirna_extension.py
uniqueg: Ah yes, good catch 👍

scripts/mirna_extension.py
uniqueg: Good point about the description. If I recall correctly, I think the previous description also said or implied that names do not change if the boundaries are not shifted (but I may be wrong). In any case, what I described is _not_ what I implemented. 

Could you please update the description and/or code according to how you want/need it and make sure they match? 

There's no indentation because the triple quotes retain whitespace. So using indentation here would probably lead to random whitespace being included in the output, depending on how `argparse` formats and wraps the description. If it bothers you, you can construct the description from multiple parts instead, e.g., like so:

```py
description = (
    "This is a very long"
    " sentence, but we care"
    " about indentation."
)
```

scripts/mirna_extension.py
uniqueg: Ah, yes, makes sense. Yes, I think it would be better to use `arguments` in the scope where you use it less.

uniqueg:
> Here are some small comments; besides the last one, everything is related to the docstrings . 
> In the meantime, if it's alright, I'll be working on fixing the unit tests and static code analyses for this PR.

Awesome, thanks a lot :)

Will you address these points when you are adding the tests etc? 🙏

deliaBlue:
> > Here are some small comments; besides the last one, everything is related to the docstrings .
> > In the meantime, if it's alright, I'll be working on fixing the unit tests and static code analyses for this PR.
> 
> Awesome, thanks a lot :)
> 
> Will you address these points when you are adding the tests etc? 🙏

On it! The unit tests will take me the rest of today, luckily not _that_ much of tomorrow.

Just a small comment. The following line has more than 80 char therefore, it fails the static code analysis:

```
precursor.attributes["Name"][0] += f"_-{pre_orig_start - precursor.start}"
```

black suggests to apply the following style:

```
precursor.attributes["Name"][
        0
    ] += f"_-{pre_orig_start - precursor.start}"
```

I find it horrible. So I want to know which one of the following approach you want me to follow:

1. Create the `suffix` variable, blending both formatted strings to afterwards modify the precursor name.
2. Change the `precursor` name to something shorter.
3. Change the `pre_orig_start` name to something shorter.
4. Use black's suggestion.

I'd rather go with the first one but it's up to you :)

deliaBlue:
The changes applied in the last push are:

- Update the unit test and the required files for the script 'mirna_extension.py'
- Refactor the 'mirna_extension.py' script
- Update expected output (only the extended GFF3 files change) to match the new script (see [this comment](https://github.com/zavolanlab/mirflowz/pull/142#discussion_r1605879792) for more information on the changes)
- Two new errors have been add to ignore (W0719 in order to use `rise Exception` and E0606 (possibly-used-before-assignment) as an error is raised prior to using that variable in the script 'mirna_quantification.py' lines 260, 304 and 330)

deliaBlue (approved):

uniqueg:
> > > Here are some small comments; besides the last one, everything is related to the docstrings .
> > > In the meantime, if it's alright, I'll be working on fixing the unit tests and static code analyses for this PR.
> > 
> > 
> > Awesome, thanks a lot :)
> > Will you address these points when you are adding the tests etc? 🙏
> 
> On it! The unit tests will take me the rest of today, luckily not _that_ much of tomorrow.
> 
> Just a small comment. The following line has more than 80 char therefore, it fails the static code analysis:
> 
> ```
> precursor.attributes["Name"][0] += f"_-{pre_orig_start - precursor.start}"
> ```
> 
> black suggests to apply the following style:
> 
> ```
> precursor.attributes["Name"][
>         0
>     ] += f"_-{pre_orig_start - precursor.start}"
> ```
> 
> I find it horrible. So I want to know which one of the following approach you want me to follow:
> 
> 1. Create the `suffix` variable, blending both formatted strings to afterwards modify the precursor name.
> 2. Change the `precursor` name to something shorter.
> 3. Change the `pre_orig_start` name to something shorter.
> 4. Use black's suggestion.
> 
> I'd rather go with the first one but it's up to you :)

Either way is fine by me :)

scripts/mirna_extension.py
```diff
@@ -1,240 +1,375 @@
...
+
+                    try:
+                        max_len = anno_lengths[ref_seq]
+                    except KeyError:
+                        max_len = 0
+
+                    try:
+                        self.seq_lengths[ref_seq] = int(length)
+                        if max_len > int(length):
+                            raise Exception(
```
uniqueg, line 73: Why not raise a more specific exception. You could, e.g., create a custom exception `AnnotationException` that inherits from `ValueError`.

However, note that this statement should be outside of the `try`-`except` block - which is meant to only catch integer conversion errors. In any case it does not make sense to be inside `try`-`catch`, as the `if` clause acts as a replacement for `try`-`catch`.

pylint.cfg
```diff
@@ -1,2 +1,2 @@
 [MESSAGES CONTROL]
-disable=C0103,E1101,I1101,R0801
+disable=C0103,E1101,I1101,R0801,E0606,W0719
```
uniqueg, line 2: Remove `W0719` (see comment in script below). If you absolutely want to catch broad `Exception`s, ignore the issue in line, not globally in the Pylint config. It is really not good practice to catch broad exceptions and therefore should only be used in rare cases. Disabling the check globally therefore invites bad practice.

uniqueg, line 2: Also please remove `E0606`. I have read why you added it but unfortunately don't find the variable in question as the line numbers you give don't seem to match (probably outdated).

In any case, there is almost _always_ a better way to ensure that `E0606` is not raised. And even if not, I still find it better to disable it per line or per block, not globally. In the config we only want to put rules that we generally disagree with, in (almost) all cases.

uniqueg, line 2: Btw, in principle this goes for the other rules as well. Neither of the listed rules should need to be disabled globally, I think...

scripts/tests/files/extreme_chr_mir_anno.gff3
```diff
@@ -1,3 +1,2 @@
-19	.	miRNA	0	80	.	+	.	ID=MIMAT0002822;Alias=MIMAT0002822;Name=hsa-miR-512-5p;Derives_from=MI0003140
-19	.	miRNA	83	124	.	+	.	ID=MIMAT0002823;Alias=MIMAT0002823;Name=hsa-miR-512-3p;Derives_from=MI0003140
+19	.	miRNA	1	80	.	+	.	ID=MIMAT0002822;Alias=MIMAT0002822;Name=hsa-miR-512-1-5p;Derives_from=MI0003140
```
uniqueg, line 1: By setting this to 1 (which is legal) instead of 0 (which is illegal) here an in the next file - are you still testing what happens when a GFF annotation has a 0 (illegal) coordinate?

uniqueg (commented):
A few comments...

pylint.cfg
deliaBlue: Should I open an issue for this? Or can I fix it here?

deliaBlue: I put it globally because it was affecting other scripts and as that particular script will be modified in another PR I though of taking this shortcut to avoid blocking everything, but I can fix it here directly as the original PR was to fix the static code analysis.

uniqueg: Let's just leave the rules that are there and only remove the new ones. I don't think this needs an issue, it's not important enough. Was more of a general comment/advice to think twice when globally disabling linting rules.

uniqueg: While `E0606` can give false positives, using variables before assignment is always a critical error - so if you disable this check globally, it becomes much easier to introduce such bugs accidentally. Quite a price to pay - especially since usually it is very easy to avoid the warning. If you tell me the code in question, I can give you advice on how to avoid it.

scripts/tests/files/extreme_chr_mir_anno.gff3
deliaBlue: I added a `try-except` block in the local database creation method `set_db` that raises the custom annotation exception and its correspondent unit test and file. This way, as we are not giving the possibility of setting a starting position at 0, if the initial file is wrong, the whole script stops.

deliaBlue:
The pushed changes are:

- Create custom exception to accurately catch and raise annotation errors
- Account for possible illegal coordinates in the user-provided GFF3 file (add `try-except` block + one unit test + one unit test file)
- Remove added errors to ignore in `pylint.cfg`
- fix minor static code analysis errors in `mirna_quantification.py` and in `validation_fasta.py`

uniqueg:
LGTM :)

PR #146, opened 2024-08-18
fix: classify mature miRNAs

### LINKED ISSUE #143: fix: classify correctly mature miRNA
The current naming strategy has two problems:

1. The usage of the CIGAR and MD strings does not provide a name for isomiRs with a 1-to-1 relation between the name the sequence it represents. Therefore, the results can contain 1-to-many relations.
2. The start and end position shifts are not computed correctly creating incorrect names and possibly swapping canonical miRNA and isomiRs.

In order to solve this issue, the script `iso_name_tagging.py` needs to be refactor knowing exactly the values (if 0-based or 1-based) of the start and end positions in both the input files and the tools used. 
In addition, find a new naming convention that ensure 1-to-1 relationships between the isomiR and the sequence it represents and apply it in that same script.

---

During the creation and review of this issue PR (PR #146) several extra tasks appeared and was getting too big. In order to properly tackle everything, three different sub-issues are going to be used.

deliaBlue:
This PR closes #143 .

The isomiR notation used until now was not unambiguous and lead to incorrect counts provided that the CIGAR and MD strings could be the same for different isomiR sequences. To account for this fact, the read sequence is added to the isomiR name.

The changes required to accomplish that are:

- Modify the `iso_name_tagging.py` script to add the read sequence on the isomiR name
- Modify the unit test for the script `iso_name_tagging.py` to account for the isomiR new name format, and the corresponding files. Given that the previous unit test file was not testing for the functions alone, those tests have been added.
- Modify the `mirna_quantification.py` script to account for the new name format. 
- Modify the unit tests for the script `mirna_quantification.py` to account for the new name format, and the corresponding files.
- Update the expected output of the pipeline to include the read sequence on the isomiR names

In addition and to generalize the script scope:

- Add new CLI argument in `iso_name_tagging.py` (`--shift`)
- Document `iso_name_tagging.py` in a more general way
- Modify unit tests to account for the new CLI argument
- Add new argument to the `quantify.smk` workflow
- Rename `iso_name_tagging.py` to `annotate_sam_with_bed_features.py`


scripts/iso_name_tagging.py
```diff
@@ -37,9 +37,9 @@
     command:
         iso_name_tagging.py -b BED -s SAM
     new name:
-        hsa-miR-524-5p|0|0|11M3I11M|22
+        ""
```
uniqueg, line 40: Why did this change? Would be good to give an explanation for each example. It's not clear why no tag is added here.

Also, are you sure that an empty string doesn't cause problems? Does it not violate the SAM specs?

scripts/iso_name_tagging.py
```diff
@@ -49,9 +49,9 @@
     command:
         iso_name_tagging.py -b BED -s SAM
     new name:
-        ""
+        hsa-miR-1323|0|0|21M|21|TCAAAACTGAGGGGCATTTTC
```
uniqueg, line 52: Same as above: Not clear why this changed.

uniqueg (changes requested):
In some places, the documentation of the script is written as if the script is a general purpose script for adding intersecting features to SAM files as a tag. But in other places, it becomes clear that there are several assumptions that limit the scope of the script (e.g., `miRNA_ID`, shifts/extensions) to miRNAs and isomiRs. I think you could clarify that a bit better and perhaps also make a reference to the scripts that produce valid inputs to this file, because it is highly unlikely that someone would create the inputs for this script manually.

Btw, it would have actually been nice to design this script such that it actually _is_ a general purpose script for adding "name" tags for intersecting features to SAM files and do all the other stuff (dealing with alignments that don't have an intersecting feature, dealing with maximum extensions etc) elsewhere. As it is, this script is quite complicated and has basically zero chance of reuse outside of this workflow.

Anyway, not important now - just lessons for the future :)

So please just clarify the scope of the script in the module-level docstring and I think we are ready to go.

scripts/iso_name_tagging.py
deliaBlue: After looking at the documentation and some related issues on GitHub, I found [this answer](https://github.com/GFA-spec/GFA-spec/issues/124#issuecomment-2174511060) which I believe explains quite well  why empty strings are allowed on the optional string type tag Z (the one used in this script)

deliaBlue:
> In some places, the documentation of the script is written as if the script is a general purpose script for adding intersecting features to SAM files as a tag. But in other places, it becomes clear that there are several assumptions that limit the scope of the script (e.g., `miRNA_ID`, shifts/extensions) to miRNAs and isomiRs. I think you could clarify that a bit better and perhaps also make a reference to the scripts that produce valid inputs to this file, because it is highly unlikely that someone would create the inputs for this script manually.
> 
> Btw, it would have actually been nice to design this script such that it actually _is_ a general purpose script for adding "name" tags for intersecting features to SAM files and do all the other stuff (dealing with alignments that don't have an intersecting feature, dealing with maximum extensions etc) elsewhere. As it is, this script is quite complicated and has basically zero chance of reuse outside of this workflow.
> 
> Anyway, not important now - just lessons for the future :)
> 
> So please just clarify the scope of the script in the module-level docstring and I think we are ready to go.

I believe the script _per se_ is pretty general: it adds a custom tag showing which features an alignment intersects with. I think the problem comes when defining variables and how it is documented. In this sense, the script appears to be only for miRNAs whose annotations might or might not been previously extended. But if the word `extension` is changed to `shift` or `range` and the description changes to "Allowed shift range between either end of the alignment and the intersecting feature" (or maybe a better description but just for you to get the idea) then it gets more general and the script does not have to change that much.
Another example would be the way the tag format is specified. If instead of using `miRNA_ID` I use `intersecting_feature` any kind of sequence can be used as long as it has been intersected using Bedtools intersect.

I suggest to try and make the descriptions and names more general and if you do not see it clear, I will just revert the commit and document a more restricted scope.

uniqueg:
Yes, you can do that if you like. The shift stuff is still quite specific, but I think it's best to keep it, so that we can come to an end on this soon and publish the workflow :)

uniqueg (changes requested):
Apart from the documentation issues, it should be fine.

scripts/iso_name_tagging.py
```diff
@@ -6,27 +6,25 @@
...
+FEATURE_ID|5'-shift|3'-shift|CIGAR|MD|READ_SEQ.
 
 EXPECTED INPUT FILES
 The BED file must be the output of a bedtools intersect call with `-a` being a
-GFF3 file and `-b` a BAM file. If the GFF3 used in the bedtools intersect call
-has the features start and end coordinates extended, the number of additional
-nucleotides must be specified using the CLI option `--extension`. The SAM file
-must contain only the reads that have an intersecting feature.
+GFF3 file and `-b` a BAM file. The SAM file must only
+contain alignments with an intersecting feature. If either the BED or the SAM
```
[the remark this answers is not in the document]
deliaBlue: I will add a `try - except` block in order to account for alignments in the input SAM file without an intercepting feature. This way there won't be complications in the input SAM format nor the description of it

scripts/tests/test_annotate_sam_with_bed_features.py
```diff
@@ -452,14 +452,14 @@ def test_main_bed_sam_file(self, monkeypatch, capsys, bed_sam):
     def test_main_bed_sam_extension_file(
         self, monkeypatch, capsys, bed_sam_extension
     ):
-        """Test main function with extension and allowed shit equal to 6."""
+        """Test main function with extension and allowed shift equal to 6."""
```
uniqueg, line 455: :rofl:

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+    """
+    cigar = alignment.cigarstring
+    seq = alignment.query_sequence
+    md = alignment.get_tag("MD")
+
+    tags = []
+
+    for feat_name, feat_start, feat_end in intersecting_feat:
+        shift_5 = alignment.reference_start - feat_start + 1
+        shift_3 = alignment.reference_start + alignment.query_length - feat_end
```
uniqueg, line 285: Are you sure that `alignment.query_length` is correct here? What if there are insertions or deletions in the query? Wouldn't that throw off the position relative to the reference?

Wouldn't this do the trick instead?

```py
shift_3 = alignment.reference_end - feat_end
```

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+        if -shift <= shift_5 <= shift and -shift <= shift_3 <= shift:
+            tags.append(f"{feat_name}|{shift_5}|{shift_3}|{cigar}|{md}|{seq}")
+
+    return set(tags)
```
uniqueg, line 289: Do you really need a set here? Doesn't the code already ensure that the values in tags are unique?

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+
+Optional adjustments:
+--extension: Adjust the feature's start and end coordinates by the given value
+    (start is increased and end decreased).
+--shift: Requires both the 5' and 3' shift values to be within +/- this value
+    to include the feature tag in the alignment.
+
+If an alignment has multiple intersecting features, the tag values are
+concatenated using a semicolon as the separator. If no valid intersecting
+feature is found, the alignment is skipped.
```
uniqueg, line 34: Is this last sentence accurate? It seems to me that the alignment is only skipped if there are no intersecting features _or_ if none of the intersecting features pass the shift filter.

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+    else:
+        attr_dict = {
+            p.split('"')[0].strip().lower(): p.split('"')[1] for p in pairs
+        }
+
+    return attr_dict
+
+
+def parse_intersect_output(
+    intersect_file: Path, ID: str = "name", extension: int = 0
```
uniqueg, line 187: `ID` is a bad choice for a variable name, because:
- it violates the convention that variables are lower case and constants are upper case
- worse: it masks the built-in `id()` function

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+        intersecting_feat: list of tuples (feature identifier, start and end).
+        alignment: A pysam.AlignedSegment object
+        shift: Maximum allowed absolute shift for both ends (default: 0).
+
+    Returns:
+        A set of tag strings constructed from valid intersecting features.
+    """
+    cigar = alignment.cigarstring
+    seq = alignment.query_sequence
+    md = alignment.get_tag("MD")
```
uniqueg, line 279: What if it isn't there?

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+            "read_strand",
+            "overlap_len",
+        ),
+    )
+
+    with open(intersect_file, "r", encoding="utf-8") as bedfile:
+        for line in bedfile:
+            fields = Fields(*line.strip().split("\t"))
+
+            feat_name = attributes_dictionary(fields.feat_attributes)[ID]
```
uniqueg, line 238: What if the `ID` attribute doesn't exist?

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+            "feat_attributes",
+            "read_chr",
+            "read_start",
+            "read_end",
+            "read_name",
+            "read_score",
+            "read_strand",
+            "overlap_len",
+        ),
+    )
```
uniqueg, line 232: Is this really a BED file? My bioinfo knowledge is a bit rusty, but I thought BED didn't have score, but rather had strand within the first 6 fields. It rather looks like GTF to me.

Also, what's the behavior if your "BED" file doesn't conform to the expected format (which, IMO, is 100% of the cases, because it's not BED at all - or maybe I'm missing sth)? I think you should properly validate it. Also, might be a good idea to use sth like BioPython to check if you are actually working with a valid extension of BED (or GTF) and not just some custom format.

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+    BED record:
+        19	.	miRNA	5332	5365	.	+	.	ID=MIMAT0005795;Alias=MIMAT0005795;Name=hsa-miR-1323;Derives_from=MI0003786	19	5337	5358	48-1_1	255	+	21
+    SAM record:
+        48-1_1	0	19	5338	255	21M	*	0	0	TCAAAACTGAGGGGCATTTTC	*	MD:Z:21	NH:i:1	NM:i:0
+    Command:
+        annotate_sam_with_bed_features.py -b BED -s SAM --extension 6 --shift 7
+    Outcome:
+        After adjusting the feature coordinates, both the 5' and 3' shifts are
+        within the allowed range, so the tag
+        "hsa-miR-1323|0|0|21M|21|TCAAAACTGAGGGGCATTTTC" is added.
```
uniqueg, line 70: If there is a shift, shouldn't at least one of the numbers _not_ be a zero?

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+        A set of tag strings constructed from valid intersecting features.
+    """
+    cigar = alignment.cigarstring
+    seq = alignment.query_sequence
+    md = alignment.get_tag("MD")
+
+    tags = []
+
+    for feat_name, feat_start, feat_end in intersecting_feat:
+        shift_5 = alignment.reference_start - feat_start + 1
```
uniqueg, line 284: Are you 100% sure the offsets (0 vs 1) are properly considered, because this is always such a hassle with SAM being 1-based, BED not being 1-based, pysam being whatever-based... And the somewhat complex handling with shifts and extensions doesn't really help.

BTW, I think this script really needs alignments in the examples, so that it becomes perfectly clear what is what.

scripts/annotate_sam_with_bed_features.py
```diff
@@ -0,0 +1,328 @@
...
+    BED record:
+        19	.	miRNA	5338	5359	.	+	.	ID=MIMAT0005795;Alias=MIMAT0005795;Name=hsa-miR-1323;Derives_from=MI0003786	19	5337	5358	48-1_1	255	+	21
+    SAM record:
+        48-1_1	0	19	5338	255	21M	*	0	0	TCAAAACTGAGGGGCATTTTC	*	MD:Z:21	NH:i:1	NM:i:0
+    Command:
+        annotate_sam_with_bed_features.py -b BED -s SAM
+    Outcome:
+        The tag "hsa-miR-1323|0|0|21M|21|TCAAAACTGAGGGGCATTTTC" is appended
+        because the alignment perfectly matches the feature coordinates.
+
```
uniqueg, line 59: I think examples would be useful that (1) use only `--shift` and (2) use only `--extension`. It really doesn't help clarity that the available examples use either none or both.

Also, I think the examples would benefit from:
- Stating the use case in words (to clarify the use of `--shift`, `--extensions`) etc.
- Follow the use case with the command.
- _Then_ show the BED and SAM records.
- Then add an actual alignment to visualize what is happening here.
- Add the actual output record.
- Finally add the verbal description/reasoning for the output.

uniqueg (changes requested):
There are still quite a number of issues, including potential critical ones (possibly wrong calculation of `shift_3`). I think this, together with the huge documentation that is still not sufficient, is very good evidence that the script is way too complex. I'm not suggesting to break it up at this point - but you absolutely need to be/make sure that what you do is 100% correct and properly and unambiguously documented.

scripts/annotate_sam_with_bed_features.py
deliaBlue: I've been checking the [bedtools interesect documentation](https://bedtools.readthedocs.io/en/latest/content/tools/intersect.html) and the call in the workflow. In the call we have the `-bed` parameter. This lead to the conclusion that the output file is a BED file. But the description of this parameter is:

> When using BAM input (-abam), write output as BED. The default is to write output in BAM when using -abam. For example: bedtools intersect -abam reads.bam -b genes.bed -bed.

We use -`-a` for the GFF3 file and `-b` for the BAM one. We are not using the param `-abam` so `-bed` does nothing.

Two parameters which we use are `-wb` and `-wo` to intersect pre-miRs and miRs respectively.:
- `-wb` writes the original A and B entries (_i.e._ the GFF3 and BAM entries in the same line).
- `-wo` does the same as `-wb` but adds the number of basepairs of overlap between both entries

That means that we have a custom format:

GFF3 format + BED6 format ( + int )


Given that for Oligomap we use the extension `.oligmap` in this case, we could use `.intersect` and check in the script that the input format is the one expected.

scripts/annotate_sam_with_bed_features.py
deliaBlue: After checking the format of the input file (see previous comment), I can better clarify the way the different offsets are accounted for:

- The format of the "BED" input file is: GFF3 + BED6 + INT. This means that the features coordinates are 1-based (as it is expected from a GFF3 file)
- The SAM input file, is 1-based BUT `pysam` is 0-based (as you can see at the first note on[ their documentation](https://pysam.readthedocs.io/en/latest/api.html#introduction))

Therefore, I can surely state that there is an offset between coordinates to be accounted for.

I will add visual examples and a better explanation on the docstrings to make this more clear.

scripts/annotate_sam_with_bed_features.py
uniqueg: If GFF3 allows extra columns, it is still a valid GFF3 file, I guess, not entirely a custom format.

Importantly though, what it is clearly _not_ is a valid BED file - which the docstring, variable names etc. all claim it is.

So please just make sure this is all consistent:
- If it's a valid GFF3 file (please check!), give it the appropriate ending
- If it is NOT a valid GFF3 file, you can use `.intersect` as you suggest

EITHER WAY: Update the docstring, variable names etc to be perfectly clear what kind of beast you are expecting here, _especially_ if you are using any of the non-standard GFF3 fields in the script (which I think you do).

I think it would also make sense to validate the format within the script and ensure that it behaves gracefully when someone throws a non-valid input file at it. In particular, run the script with any old GFF3 or BED file as inputs. It should do the most reasonable thing possible in this case. Maybe it can still do its job (but without some special features), or maybe (more likely) the best way is to refuse running with an informative error message, like:

> You have passed an invalid input file. Make sure your input file was generated with ... Open the documentation with `script_name.py -h` to learn more.

scripts/annotate_sam_with_bed_features.py
deliaBlue: Well, it is indeed a bit more complex than that.  To clarify the reasoning, here’s a breakdown of the relevant fields and their behavior:

- [`reference_start`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.reference_start) is the  0-based leftmost coordinate

- [`reference_end`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.reference_end) points to one past the last aligned residue

- [`query_length`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.query_length) corresponds to the length of the sequence supplied in the BAM/SAM file (including matches and InDels).


Now, let us consider the following alignment:

```console
# SAM ENTRY
read_1	0	19	44414	255	21M	*	0	0	GAAGGCGCTTCCCTTTGGAGT	*	MD:Z:21	NH:i:1	NM:i:0

# VISUAL REPRESENTATION
GAAGGCGCTTCCCTTTGGAGT    ( Ref. sequence )
 | |  |  |  |  | |  | |  | |  |  |  | | |  |  | | |  |     ( 21 matches )
GAAGGCGCTTCCCTTTGGAGT    ( Read sequence )
```

In this case:
- `alignment.reference_start` = 44413
- `alignment.query_length` = 21
- `alignment.reference_end` = 44434
- `alignment.reference_start` + `alignment.query_length` = 44434

So here, `alignment.reference_end` matches `alignment.reference_start + alignment.query_length`, which could lead one to think it's always safe to use `alignment.reference_end` to get the read's extent.


But what happens when there are InDels in the alignments?

```console
# SAM ENTRY
read_2	0	19	7626	1	12M3I11M	*	0	0	AAAGCACCTCCATTTGAGCTTGAAGC	*	HI:i:0	MD:Z:23	NH:i:1	NM:i:3	RG:Z:A1	YZ:Z:0

# VISUAL REPRESENTATION
AAAGCACCTCCA - - - GAGCTTGAAGC   ( Ref. sequence ) 
 | |  |  |  |  |  | | |  | |  |         |  | | |  |  | |  |  | |  |    ( 3 Insertions )
AAAGCACCTCCATTTGAGCTTGAAGC   ( Read sequence )
```


In this case:
- `alignment.reference_start` = 7625
- `alignment.query_length` = 26
- `alignment.reference_end` = 7648
- `alignment.reference_start` + `alignment.query_length` = 7651

Here, `alignment.reference_end` is less than `alignment.reference_start + alignment.query_length`, because the inserted bases ( `'TTT'` ) are not part of the reference. Thus, `alignment.reference_end` only tells us how far the alignment extends on the reference, not how long the read actually spans.

---

When using tools like `bedtools intersect`, we're comparing intervals. Our goal is to check whether a genomic feature **overlaps with the entire read, not just its aligned portion**. Since `reference_end` ignores InDels, it may underestimate the span of the read in cases with InDels.

To properly account for the full read when testing for overlap (especially when full read coverage is important), we should compute the interval as:

```python
alignment.reference_start + alignment.query_length
```

This ensures that all bases of the read — aligned, deleted or inserted — are considered when determining overlaps.

scripts/annotate_sam_with_bed_features.py
deliaBlue: As far as I can tell, the GFF3 format does not support additional columns, so we are effectively working with a custom format.

In addition to the updates you mentioned, I will include input file validation in this script, but I'm currently debating whether to validate each line individually within the `parse_intersect_output()` function—either by integrating the logic directly or calling a separate helper function—or to implement a completely separate function that validates the entire input file before passing it to `parse_intersect_output()`.

While a separate validation function would make the code cleaner and more modular, validating line by line during parsing would likely be more efficient. @uniqueg, what do you prefer in this specific case?

scripts/annotate_sam_with_bed_features.py
uniqueg: Thanks for the detailed explanation. Make sure that you update the documentation in the script so that it explains this as well as possible. No need to go into _that_ level of detail, but users should be well aware of what exactly is happening. You can resolve this issue once you have updated the docs accordingly (if an update is necessary, that is).

scripts/annotate_sam_with_bed_features.py
uniqueg: I'm pretty sure GFF3 does _not_ forbid additional columns (column 10, 11 etc.). But you can check and see if GFF parsers (BioPython, gffutils, R) complain about it. Also, column 9 of GFF3 allows specifying arbitrary key/value pairs. I don't know which format is produced by the previous call (custom fields in 9th column as KV pairs OR custom fields in 10th, 11th etc. columns), but I think either of these are still GFF3 compliant.

Re: validation, I would do the following:
- Fully validate the first 10 lines before passing the file
- For each line, use regular exception handling to validate that the line has what you need

This will be sufficient to cover MOST cases. If there is a file that starts to be a valid input file and then somehow changes to records that retain the general _shape_ of a valid input file (evidenced by the presence and correct type of the columns we need to look at), but _actually_ contains nonsense, then it is really not our fault if the script produces a "wrong" output.

deliaBlue: Also, I just realized that the script [`primir_quantification.py` ](https://github.com/zavolanlab/mirflowz/blob/dev/scripts/primir_quantification.py) *also* receives as input file the outcome of a bedTools intersect call. Should I add the same validation and documentation as in this one?

This PR is getting huge but the validation function, the change in the file extension and documentation is also affecting this script.

deliaBlue:
I'm closing this PR as it has been broken down into several steps as explained in https://github.com/zavolanlab/mirflowz/issues/143 and it is not going to be merged.

PR #150, opened 2025-01-19
ci: update CI test file

deliaBlue:
This PR update the CI test file by:

- Upgrade the miniconda version to v3
- Pin the ubuntu version to `20.04`
- Add `$CONDA_PREFIX` in the environment updating
- Add `auto-update-conda: false` and `channel-priority: true`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ on:
 jobs:
 
   static-code-analysis:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     defaults:
       run:
         shell: bash -l {0}
@@ -21,15 +21,17 @@ jobs:
         uses: actions/checkout@v4
          
       - name: Setup Conda/Mamba
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
         with:
           mamba-version: "*"
+          channel-priority: true
+          auto-update-conda: false
           activate-environment: mirflowz
           environment-file: environment.yml
           auto-activate-base: false
       
       - name: Update mirflowz env with dev packages
-        run: mamba env update -n mirflowz -f environment.dev.yml
+        run: mamba env update -p $CONDA_PREFIX -n mirflowz -f environment.dev.yml
       
       - name: display environment info
         run: |
@@ -53,7 +55,7 @@ jobs:
         run: black --check --verbose --line-length=79 ./*.py
 
   snakemake-format-graph-test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     defaults:
       run:
         shell: bash -l {0}
@@ -64,15 +66,17 @@ jobs:
         uses: actions/checkout@v4
          
       - name: setup Conda/Mamba
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
         with:
           mamba-version: "*"
+          channel-priority: true
+          auto-update-conda: false
           activate-environment: mirflowz
           environment-file: environment.yml
           auto-activate-base: false
       
       - name: update mirflowz env with dev packages
-        run: mamba env update -n mirflowz -f environment.dev.yml
+        run: mamba env update -p $CONDA_PREFIX -n mirflowz -f environment.dev.yml
       
       - name: display environment info
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
 
   static-code-analysis:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
uniqueg, line 14: Ubuntu 20-04 is deprecated since the beginning of the month: https://github.com/actions/runner-images/issues/11101

Please update to `ubuntu-24.04` - or better yet: revert back to `ubuntu-latest` (it's not quite clear to me why you pinned/downgraded it)

.github/workflows/tests.yml
```diff
@@ -21,15 +21,18 @@ jobs:
...
+        uses: conda-incubator/setup-miniconda@v3
         with:
-          mamba-version: "*"
+          mamba-version: "2"
+          channel-priority: true
+          auto-update-conda: false
           activate-environment: mirflowz
           environment-file: environment.yml
+          conda-remove-defaults: true
           auto-activate-base: false
```
uniqueg, line 32: I think it would make sense to always specify the arguments in the same order whenever you use the `setup-miniconda` action. If it doesn't break anything (which I guess/hope it shouldn't), perhaps the best is to use increasing alphabetic order.

.github/workflows/tests.yml
```diff
@@ -120,11 +129,13 @@ jobs:
         run: bash test/test_workflow_local_with_singularity.sh
 
       - name: run local test with Conda
-        run: bash test/test_workflow_local_with_conda.sh
+        run: |
+          conda config --set channel_priority strict
```
uniqueg, line 133: What's the reason behind changing the channel priority here? If I see this right, one set of tests runs with `true` and the other with `strict`, right? Why?

uniqueg (changes requested):
As there is no issue attached to this PR, I'm not quite sure what's the significance of this PR. Sure, it's good to update the dependencies, but was there anything in particular that prompted you to do so? No problem if not, just curious.

Apart from that, I'm not sure why you pinned the Ubuntu images, but if you really need to pin them, we will need Ubuntu 24.04, given that Ubuntu 20.04 is now deprecated and will soon be unsupported.

A few other minor issues, should only take a few min.

uniqueg:
The PR looks good from what I can see. If you fix the integration tests, I think we can merge :)

deliaBlue:
Now that the singularity CI test has been removed, an old error has appeared. 

When running the pipeline using Conda and when using Singularity the output content remains the same but the order in which it is displayed within the file changes (see `diff` output bellow) dragging the same error across several files. Given the nature of the md5sum test, the check fails. By now, I'll update the expected output and approach this error in the [Singularity issue](#151).


```
$diff singularity.out conda.out

8d7
< 19	.	miRNA	9080	9113	.	+	.	ID=MIMAT0037324;Alias=MIMAT0037324;Name=hsa-miR-520e-5p;Derives_from=MI0003143
9a9
> 19	.	miRNA	9080	9113	.	+	.	ID=MIMAT0037324;Alias=MIMAT0037324;Name=hsa-miR-520e-5p;Derives_from=MI0003143
14d13
< 19	.	miRNA	15527	15560	.	+	.	ID=MIMAT0026609;Alias=MIMAT0026609;Name=hsa-miR-520f-5p;Derives_from=MI0003146
15a15
> 19	.	miRNA	15527	15560	.	+	.	ID=MIMAT0026609;Alias=MIMAT0026609;Name=hsa-miR-520f-5p;Derives_from=MI0003146
33d32
< 19	.	miRNA	34582	34614	.	+	.	ID=MIMAT0037325;Alias=MIMAT0037325;Name=hsa-miR-520b-5p;Derives_from=MI0003155
34a34
> 19	.	miRNA	34582	34614	.	+	.	ID=MIMAT0037325;Alias=MIMAT0037325;Name=hsa-miR-520b-5p;Derives_from=MI0003155
46d45
< 19	.	miRNA	46715	46751	.	+	.	ID=MIMAT0026610;Alias=MIMAT0026610;Name=hsa-miR-519d-5p;Derives_from=MI0003162
47a47
> 19	.	miRNA	46715	46751	.	+	.	ID=MIMAT0026610;Alias=MIMAT0026610;Name=hsa-miR-519d-5p;Derives_from=MI0003162
```

uniqueg (approved):
LGTM

Issue #164, opened 2025-08-18
fix: classify mature miRNA


deliaBlue:
The isomiR notation used until now was not unambiguous and lead to incorrect counts provided that the CIGAR and MD strings could be the same for different isomiR sequences. To account for this fact, the read sequence is added to the isomiR name.

### What needs to be done 

- Modify the `iso_name_tagging.py` script to add the read sequence on the isomiR name.
- Modify the `mirna_quantification.py` script to account for the new name format.
- Modify the unit test and unit test files for the two scripts to account for the new name format.

deliaBlue:
This issue was solved in https://github.com/zavolanlab/mirflowz/commit/3e3750e3dbc736cb7b97d1f94370345293804291

Issue #163, opened 2025-08-18
refactor: generalize `iso_name_tagging.py`


deliaBlue:
Right now, the script [iso_name_tagging.py](https://github.com/zavolanlab/mirflowz/blob/dev/scripts/iso_name_tagging.py) has a very specific scope and a very confusing documentation.

### What needs to be done

- Generalize the script and clarify its scope
- Give it an appropriate name
- Properly document it (See  https://github.com/zavolanlab/mirflowz/pull/146#discussion_r2004022470)

PR #168, opened 2025-09-18
feat: parse and validate intersect custom format

### LINKED ISSUE #162: feat: parse and validate custom format
The output of `bedtools intersect` has been wrongly assigned as a BED file. 

Two different scripts ([iso_name_tagging.py](https://github.com/zavolanlab/mirflowz/blob/dev/scripts/iso_name_tagging.py) and [primir_quantificaion.py](https://github.com/zavolanlab/mirflowz/blob/dev/scripts/primir_quantification.py)) read and process this kind of file.

### What needs to be done

- Add a file format validation in both scripts
- Add unit tests for these validation for both scripts
- Change all docstrings, file extensions and CLI flags from `bed` to `intersect`

deliaBlue:
This PR closes #162.

Previously, the pipeline  had two different calls, now both of them are the same. Hence,  A single validation module for the `bedtools intersect -wo -s` output is needed and has been created.

The `iso_name_tagging.py` script has a lot of things to be changed (see #163 and #164), therefore, in this PR only the inclusion of the validation module must be reviewed.

The expected output has been modified due to:
1. Files having a different extension (BED -> INTERSECT)
2. The pri-miR intersection now having the overlap column

Besides that, everything remained the same.

## Summary by Sourcery

Add a unified validation and parsing module for bedtools intersect -wo -s output, integrate it into the existing quantification and tagging scripts, and update file extensions from .bed to .intersect throughout the codebase, tests, documentation, and workflow.

New Features:
- Introduce validate_bedtools_intersect module with Record, FileFormatError, validate_first_n, and parse_all to strictly parse and validate intersect output.
- Integrate the validation module into primir_quantification.py and iso_name_tagging.py to enforce input format checks before processing.
- Standardize intersection file extension from .bed to .intersect across scripts, tests, documentation, and Snakemake rules.

Enhancements:
- Remove ad-hoc parsing code in scripts in favor of the shared validation API.
- Bump script versions: primir_quantification to 1.1.0 and iso_name_tagging to 1.2.0.

Tests:
- Update and expand test fixtures and assertions in test scripts to use .intersect files and cover validation error cases.


uniqueg:
I don't understand this:

> The iso_name_tagging.py script has a lot of things to be changed (see https://github.com/zavolanlab/mirflowz/issues/163 and https://github.com/zavolanlab/mirflowz/issues/164), therefore, in this PR only the inclusion of the validation module must be reviewed.

If the changes shouldn't be reviewed, then they also shouldn't be part of this PR. If you need to merge a different PR first, then please raise that and let's review and merge that first, then you can merge the changes into this one, so that this only deals with what it says it deals with.

uniqueg (changes requested):
Please address https://github.com/zavolanlab/mirflowz/pull/168#issuecomment-3346278358 first

scripts/tests/test_iso_name_tagging.py
```diff
@@ -41,13 +46,22 @@ def intersect_sam_extension():
...
     return intersect_file, sam_file, output_file
 
 
+@pytest.fixture
+def invalid_file():
+    """Import path to test files with an intersect invalid line."""
+    intersect_file = Path("files/invalid_8_lines.intersect")
+    sam_file = Path("files/in_alignments_mirna.sam")
+
+    return intersect_file, sam_file
```
uniqueg: This is a bit of a weird fixture, as it hardcodes both files and is only used once. So while you save one line in your test, this is offset by using 6 lines for the fixture definition. And what if you want to use another invalid file in the tests? Write a different fixture for that?

scripts/tests/test_iso_name_tagging.py
```diff
@@ -251,3 +265,64 @@ def test_main_intersect_sam_file_id(
 
         with open(output, "r") as out_file:
             assert captured.out == out_file.read()
+
+    def test_main_invalid_intersect_file(
```
uniqueg: Rather than the fixture, you could use parametrization and use this test for both the invalid 8 and the invalid 13 lines files...

pipeline_documentation.md
uniqueg, line 1: Is it true that only the docs for `primir_quantification.py` change considerably, even though also `iso_name_tagging.py` was changed? Just checking to make sure the other was not forgotten.

scripts/validate_bedtools_intersect.py
```diff
@@ -0,0 +1,481 @@
...
+        Returns:
+            A validated 'Record' instance.
+
+        Raises:
+            FileFormatError: If the column count is wrong or any field is
+                invalid.
+        """
+        parts = line.strip().split(sep)
+
+        if len(parts) != 16:
```
uniqueg: Better to define a constant:

```py
EXPECTED_COLUMNS = 16
```

Then use that whenever you refer to the number of columns or the last column.

This makes it much easier to update if the format ever changes.

scripts/validate_bedtools_intersect.py
```diff
@@ -0,0 +1,481 @@
...
+        Ensures:
+        - Feature and read are on the same chromosome and strand.
+        - start <= end for both feature and read.
+        - Overlap > 0 (required by `-wo`) and equals the computed overlap.
+
+        Raises:
+            FileFormatError: If any rule is violated.
+        """
+        if self.feat_chr != self.read_chr:
+            raise FileFormatError(
```
uniqueg, line 236: Not saying you should change that, but defining a base error and then more specific errors for different types would be more informative/cleaner.

You could then still catch for the base `FileFormatError` in client code, but you could also respond to more specific cases, if necessary.

Since we don't need it, please leave it as is - just something for you to consider for the future :)

scripts/validate_bedtools_intersect.py
```diff
@@ -0,0 +1,481 @@
...
+        pairs = text.split(";")
+
+        if len(pairs[0].split("=")) == 2:
+            return {p.split("=")[0].lower(): p.split("=")[1] for p in pairs}
+
+        if len(pairs[0].split('"')) == 3:
+            return {
+                p.split('"')[0].strip().lower(): p.split('"')[1]
+                for p in filter(None, pairs)
+            }
```
uniqueg: This looks quite inefficient. `.split()` and `.strip()` operations can be expensive when looping over lots of records.

Better to put it all in a single loop, or probably even better yet: use a regex-based approach.

test/test_rule_graph.sh
```diff
@@ -23,4 +23,5 @@ snakemake \
     --rulegraph \
     --dryrun \
     --quiet="all" \
+| sed -n '/^\(strict \)\?digraph /,$p' \
```
uniqueg: What does this do?

scripts/iso_name_tagging.py
```diff
@@ -170,13 +171,33 @@
...
+            FileFormatError,
+            parse_all,
+            validate_first_n,
+        )
+    except ImportError:  # pragma: no cover
+        from validate_bedtools_intersect import (
+            FileFormatError,
+            parse_all,
+            validate_first_n,
+        )
```
uniqueg, line 199: I have apprehensions about this...

You import code from another module as if it were all part of a package. But it isn't packaged like a package, it's just a script, and this will not work if that script is taken out of context. And I think this is all a bit of an overkill.

After all we control how the previous step is called and so we are reasonably sure that the output follows the expected format. Moreover, the expected inputs and outputs are well documented now - so do we really have to put so much effort into validation? We aren't validating inputs/outputs in the other tools like that, do we?

Anyway, now you have done the work, let's keep it. I just hope it doesn't come around to bite us...

uniqueg (changes requested):
We need to make sure not to get into too many rabbit holes. I don't think that all of this work adds a lot to the quality of _MIRFLOWZ_.

We _could_ have built a library that contains all of the scripts as different functions and then package that, publish and containerize it etc. like a proper library. And we're kinda taking more and more steps towards that. That's all fine from a design point of view...

But it doesn't make really make _MIRFLOWZ_ any better! Rather, it it being properly releases and available. And if it's not released and not used, why worry about things like reproducibility and maintainability?

Let's try to get to an end here and keep things agile on the way. There is _always_ more to do, but let's keep the eye on the ball to get it out so that people can use it first.

All that being said, let's keep it now as you did, given that you put in all this work. But obviously I was unable to properly review all those huge changes. It's possible that bugs got introduced etc - especially because we are always just changing the `expected_output.md5` because ... well, we're changing dozens of files in a PR, so sure there are bound to be some changes in the output, right? But are we still getting the _right_ results?

So please, let's focus from now on and keep PRs small. One script changed, plus tests, with keeping changes to a minimum. No style changes suddenly (unless the PR is specifically for that), no typos, no renamings etc all put together.

pipeline_documentation.md
deliaBlue: The `iso_name_tagging.py` docstring did not change as its documentation must be addressed in #163. Just to refresh it, PR #146 was getting so huge that its issue was divided in three sub issues. The one in this PR, #163 and #164 (although #164 has nearly been completely addressed in #166.

test/test_rule_graph.sh
deliaBlue: This ensured that the messages displayed by the command above was removed. I solved the issue by removing the `--dryrun` CLI argument.

scripts/tests/test_iso_name_tagging.py
```diff
@@ -266,28 +257,25 @@ def test_main_intersect_sam_file_id(
         with open(output, "r") as out_file:
             assert captured.out == out_file.read()
 
-    def test_main_invalid_intersect_file(
-        self, monkeypatch, capsys, invalid_file
-    ):
+    @pytest.mark.parametrize("lines", ["8", "13"])
```
uniqueg: I think it's better to just enumerate the entire filenames here. Why? Imagine you change, for some reason, the filenames of the test files. Or you want to add another one. You would probably search for any instances where the filename is used. But you'd almost certainly miss it, because what you really need to search for is "8" and "13". And then, if a new or renamed file doesn't follow the naming pattern, you not only have to change the parameterization, but the actual code in the test method as well.

uniqueg (commented):
I'm not sure if it was already in a reviewable state, but went ahead and did it anyway :)

Just one minor issue left from my side. Not critical. You can leave it as is if you prefer.

uniqueg (approved):
LGTM

